PR #661, opened 2026-04-10
chore: add CI error summary step to test workflow

futuroptimist:
### Motivation
- Failures in the `CI / Run tests and collect coverage` job were hard to triage because actionable errors were buried in long logs. 
- The goal is to provide a single, concise place at the end of the job that lists every raised error so developers can find and fix failures faster. 
- The change must work even when earlier steps fail, so the summary step runs unconditionally.

### Description
- Grant the test job `actions: read` and `contents: read` permissions to allow downloading the current run logs from the Actions API by updating `.github/workflows/ci.yml`.
- Add a final `if: always()` step named `Summarize workflow errors` that downloads the workflow logs zip via `curl` and `GITHUB_RUN_ID`, unzips them, and scans the raw logs. 
- Use an inline Python parser to extract and deduplicate lines matching `::error`, `##[error]`, `Error:`, and `ERROR:` and emit a `## CI Error Summary` block. 
- Append the generated summary to the job-level GitHub summary via `$GITHUB_STEP_SUMMARY` so it appears at the end of the job output.

### Testing
- Ran `pre-commit run --all-files`, which could not execute in this environment because the `pre-commit` command is not installed (failed). 
- Ran `git diff --cached | ./scripts/scan-secrets.py`, which could not execute because `./scripts/scan-secrets.py` is not present in this container (failed). 
- The workflow changes were lint-free in-repo and the updated `.github/workflows/ci.yml` was committed locally; actual workflow behavior should be validated by pushing and observing a real GitHub Actions run to confirm the summary appears in the job output (not executed here).

------
[Codex Task](https://chatgpt.com/codex/cloud/tasks/task_e_69d89df2c6c4832fb3487e609b5c6c4c)

### DIFF
--- a/.github/workflows/ci.yml
+++ b/.github/workflows/ci.yml
@@ -65,6 +65,7 @@ jobs:
         if: always()
         env:
           GH_TOKEN: ${{ github.token }}
+          JOB_STATUS: ${{ job.status }}
         run: |
           set -uo pipefail
 
@@ -84,6 +85,7 @@ jobs:
 
           raw_dir = Path('.gha-logs/raw')
           summary_file = Path('.gha-logs/error-summary.md')
+          job_status = '${JOB_STATUS}'.strip().lower()
           ansi_re = re.compile(r'\x1B\[[0-9;]*[mGKHF]|\x1B\(B')
 
           patterns = [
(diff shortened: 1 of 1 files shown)


futuroptimist:
@codex please address the Codex, Copilot, and Greptile comments

futuroptimist:
@codex
No scope lock provided. Only touch `.github/workflows/ci.yml`; do not touch app/runtime code, tests, or unrelated workflows; keep diff small; run these commands: `python - <<'PY'\nimport yaml; yaml.safe_load(open('.github/workflows/ci.yml'))\nprint('YAML OK')\nPY` and then validate on a real GitHub Actions run for this workflow.

This is the best candidate, but it is not 100% yet. Tighten the summary so it better satisfies “lists exactly what errors were raised during the entire workflow” without expanding scope.

Required follow-up in `.github/workflows/ci.yml` only:
1. Keep the current end-of-job “download this run’s logs and summarize them” approach.
2. Broaden error extraction beyond only `::error`, `##[error]`, `Error:`, and `ERROR:`. Right now it will likely miss common failure lines from Python/pytest/npm/bash output. Update the parser to also capture at least:
   - `Traceback`
   - `Exception`
   - `FAILED`
   - `FAIL`
   - `AssertionError`
   - `fatal`
   - `ModuleNotFoundError`
   - `RuntimeError`
3. Preserve deduplication, but make sure the summary still includes the source log file for each captured line.
4. If no matching error lines are found but the workflow/job failed, emit a fallback section that lists the log files scanned and the last ~20 lines from each failing/interesting log file so the summary is never empty on a red run.
5. Do not introduce broad `continue-on-error` rewiring, step outcome plumbing, or Codecov/test gating changes in this PR. Defer any workflow-control refactor to a follow-up PR if needed.

Helpful search if you want to confirm current failure signatures in repo/tests:
- `git grep -nE 'Traceback|AssertionError|ModuleNotFoundError|RuntimeError|FAILED|\\bFAIL\\b|fatal|Error:'`
- `git grep -n "run_all_tests.sh"`
- `git grep -n "CI / Run tests and collect coverage" .github/workflows`

Verification:
1. `python - <<'PY'\nimport yaml; yaml.safe_load(open('.github/workflows/ci.yml'))\nprint('YAML OK')\nPY`
2. Push the branch and trigger the existing CI workflow on a PR.
3. Confirm the `CI / Run tests and collect coverage` job ends with a Step Summary section that:
   - appears even when an earlier step fails,
   - includes concrete error lines from the run logs,
   - cites the source log file for each line,
   - is non-empty on the known failing case from the original prompt.
4. Use the specific run from the prompt as the acceptance bar: the new summary should make the failing errors easy to find without digging through the full log archive.

If `git diff --stat` shows > 8 files or touches do-not-touch areas, stop and split.

PR #668, opened 2026-04-10
🛡️ fix: bind relay client key in encrypted legacy faucet payload

futuroptimist:
### Motivation

- The legacy `/faucet` flow trusted relay-supplied public keys for encrypting requests/responses which allowed key-substitution by a malicious relay or MITM and broke end-to-end confidentiality.

### Description

- Bind the client's public key inside the encrypted request payload sent to the relay by wrapping `chat_history` with `client_public_key` before encryption in `client.py` and `utils/crypto_helpers.py`.
- Add `_extract_chat_history_and_validate_key_binding` in `utils/networking/relay_client.py` to validate that the decrypted payload's embedded `client_public_key` matches the relay-provided metadata and reject requests if they differ, and use the validated `chat_history` for LLM calls (including streaming path).
- Add unit tests in `tests/unit/test_relay_client.py` that assert mismatched bindings are rejected and matching bindings are accepted.
- Changes touch `client.py`, `utils/crypto_helpers.py`, `utils/networking/relay_client.py`, and `tests/unit/test_relay_client.py` to implement the protection and coverage.

### Testing

- Ran `npm run lint` which completed successfully.
- Ran the project test runner via `npm run test:ci`, which failed in this execution environment due to missing/incompatible Python test dependencies (Python unit/integration test stages and crypto compatibility tests failed in-container).
- Executed `pytest -q tests/unit/test_relay_client.py` attempts but the local container initially lacked Python test dependencies (`requests`, `playwright`, `psutil`, `dotenv`, etc.), so the Python unit tests could not be executed end-to-end here; the new unit cases were added and are present for CI to run where the full Python test stack is available.

------
[Codex Task](https://chatgpt.com/codex/cloud/tasks/task_e_69d8d64ce6d0832f8c55b89dacb99386)

### DIFF
--- a/client.py
+++ b/client.py
@@ -314,7 +314,14 @@ class ChatClient:
         logger.debug("Retrieved server public key (%d bytes)", len(server_public_key))
 
         if server_public_key:
-            ciphertext_dict, cipherkey, iv = encrypt(json.dumps(self.chat_history).encode('utf-8'), server_public_key)
+            bound_payload = {
+                "chat_history": self.chat_history,
+                "client_public_key": self.public_key_b64,
+            }
+            ciphertext_dict, cipherkey, iv = encrypt(
+                json.dumps(bound_payload).encode('utf-8'),
+                server_public_key,
+            )
             encrypted_chat_history_b64 = base64.b64encode(ciphertext_dict['ciphertext']).decode('utf-8')
             iv_b64 = base64.b64encode(iv).decode('utf-8')
             encrypted_cipherkey_b64 = base64.b64encode(cipherkey).decode('utf-8')
--- a/tests/unit/test_relay_client.py
+++ b/tests/unit/test_relay_client.py
@@ -684,6 +684,49 @@ class TestRelayClient:
             timeout=relay_client._request_timeout
         )
 
+    @patch('utils.networking.relay_client.requests.post')
+    def test_process_client_request_rejects_mismatched_bound_client_key(
+        self,
+        mock_post,
+        relay_client,
+        mock_crypto_manager,
+    ):
+        """Requests are rejected when encrypted payload key binding mismatches relay key."""
+        request_data = TEST_VALID_RESPONSE.copy()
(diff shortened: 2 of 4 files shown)


futuroptimist:

futuroptimist:
@codex Thanks for the review — I’ve incorporated the requested hardening to make this merge-ready.

## ✅ Security hardening updates applied
1. **Require key-binding inside encrypted payload**
   - Relay processing now treats decrypted payloads as valid only when:
     - payload is an object,
     - `client_public_key` is present and a string,
     - it exactly matches the outer relay envelope `client_public_key`,
     - `chat_history` is present in the expected shape.
   - This removes permissive fallback behavior and closes downgrade/bypass paths.

2. **Fail-closed validation path**
   - Any malformed payload (`missing key`, `wrong type`, `mismatch`, invalid history) is rejected before LLM processing or response encryption.
   - Added explicit log messages for each reject condition to aid triage.

3. **Consistent handling for standard + streaming flows**
   - Both non-stream and stream request paths now use the same validated `chat_history` result.
   - No response is emitted to `/source` or `/stream/source` when binding validation fails.

## ✅ Test improvements
Added/updated tests to cover:
- reject on mismatched bound key,
- reject when bound key is missing,
- reject when bound key is non-string,
- accept on matching bound key,
- stream path uses validated payload only.

This gives regression coverage for both the original exploit path and common malformed variants.

## ✅ Merge-readiness checklist
- [x] Vulnerability condition addressed with fail-closed checks
- [x] Unit tests added for positive + negative cases
- [x] No secrets/tokens added
- [x] Change scope is minimal and security-focused

## Follow-up recommendation (non-blocking)
For stronger authenticity guarantees beyond relay trust:
- sign server key announcements (or pin server keys),
- verify signatures in client before encrypting to `/next_server` key material.

If reviewers are satisfied with strict fail-closed compatibility behavior, this should be good to merge.

PR #671, opened 2026-04-11
🐛 fix: surface desktop sidecar/operator startup failures

futuroptimist:
### Motivation
- The desktop Tauri host could spawn Python sidecars/bridges with `stderr` piped but never drained it, which hid errors and risked pipe backpressure making `Start operator` / `Start local inference` look like no-ops. 
- The frontend could be stranded in `starting` when an invoked command rejected because `startInference()` lacked `try/catch`, making failures invisible to users. 
- Operator/relay compatibility failures were not explicit, so an incompatible relay appeared to silently prevent registration.

### Description
- Drain and log child `stderr` for both the inference sidecar and compute-node bridge and prefix logs with grep-friendly tags, implemented in `desktop-tauri/src-tauri/src/sidecar.rs` and `desktop-tauri/src-tauri/src/compute_node.rs`. 
- Log malformed/non-JSON `stdout` lines with context instead of silently dropping them and emit an explicit UI `error` event if a child exits early without producing terminal events, implemented in `sidecar.rs` and `compute_node.rs`. 
- Make `start_inference` and `start_compute_node` return promptly by spawning the background lifecycle task and emit structured startup failure events and shell `eprintln!` logs if spawn/start fails, implemented in `desktop-tauri/src-tauri/src/main.rs`. 
- Improve frontend error wiring so `startInference()` catches invoke rejection and sets `status` to `failed`, and `startComputeNode()` surfaces `last_error` on immediate failures, implemented in `desktop-tauri/src/App.tsx`. 
- Add explicit relay compatibility messaging to the compute-node bridge for non-legacy/new protocol payloads to instruct users to upgrade the relay, implemented in `desktop-tauri/src-tauri/python/compute_node_bridge.py`. 
- Add regression tests and test infra changes: a UI unit test exercising failed local inference and failed operator start (`desktop-tauri/src/App.test.tsx`), a compute-node bridge test for incompatible relay payloads (`tests/unit/test_desktop_compute_node_bridge.py`), and comfortable Vitest setup changes in `desktop-tauri/package.json` and `desktop-tauri/vite.config.ts`. 
- Add an outage postmortem documenting symptoms, root cause and remediation (`outages/2026-04-11-desktop-tauri-operator-sidecar-silent-failure.md` and `.json`).

### Testing
- Ran frontend dependency install and unit tests with `cd desktop-tauri && npm ci` and `cd desktop-tauri && npm test`, and the Vitest suite passed (2 tests in `src/App.test.tsx` passed). 
- Ran Python unit tests with `python -m pytest tests/unit/test_desktop_inference_sidecar.py -q` and `python -m pytest tests/unit/test_desktop_compute_node_bridge.py -q`, and both test modules passed (21 and 8 tests respectively). 
- Attempted `cargo test --manifest-path desktop-tauri/src-tauri/Cargo.toml` but the environment is missing system `glib-2.0` / pkg-config which caused the Rust build to fail in this runner, so full Rust CI may require platform dev packages; the changes are formatted with `cargo fmt`. 
- `pre-commit run --all-files` was not executed in this environment because `pre-commit` is not installed here.

------
[Codex Task](https://chatgpt.com/codex/cloud/tasks/task_e_69d9e6b9eb58832f80837dda64e026fb)

### DIFF
--- a/desktop-tauri/package-lock.json
+++ b/desktop-tauri/package-lock.json
@@ -15,10 +15,179 @@
       },
       "devDependencies": {
         "@tauri-apps/cli": "^2.0.0",
+        "@testing-library/react": "^16.3.0",
         "@types/react": "^18.3.12",
         "@types/react-dom": "^18.3.1",
+        "jsdom": "^25.0.1",
         "typescript": "^5.9.3",
-        "vite": "^5.4.10"
+        "vite": "^5.4.10",
+        "vitest": "^2.1.8"
+      }
+    },
+    "node_modules/@asamuzakjp/css-color": {
+      "version": "3.2.0",
+      "resolved": "https://registry.npmjs.org/@asamuzakjp/css-color/-/css-color-3.2.0.tgz",
+      "integrity": "sha512-K1A6z8tS3XsmCMM86xoWdn7Fkdn9m6RSVtocUrJYIwZnFVkng/PvkEoWtOWmP+Scc6saYWHWZYbndEEXxl24jw==",
+      "dev": true,
+      "license": "MIT",
+      "dependencies": {
+        "@csstools/css-calc": "^2.1.3",
+        "@csstools/css-color-parser": "^3.0.9",
+        "@csstools/css-parser-algorithms": "^3.0.4",
+        "@csstools/css-tokenizer": "^3.0.3",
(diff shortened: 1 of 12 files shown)


futuroptimist:

PR #702, opened 2026-04-12
Fix desktop model-bridge PYTHONPATH bootstrap for exe/python launches

futuroptimist:
### Motivation

- The desktop `model_bridge.py` could be launched from an `exe/python` layout where shared modules live under a neighboring `resources` (or `Resources`) directory, causing `ModuleNotFoundError: No module named 'utils'` at runtime. 
- CI remained green because existing compute-node/sidecar launch paths already configured `PYTHONPATH`, while the model-bridge path did not exercise the same runtime import-root wiring.
- The intent is a minimal, targeted fix that reuses the repo-root import-root discovery and ensures the real desktop launch layout is covered by tests.

### Description

- Extend `desktop-tauri/src-tauri/python/path_bootstrap.py` to consider additional import-root candidates for `exe/python` launches by adding `resources`, `Resources`, and preserving bundled/script roots when resolving runtime import paths. 
- Add `repo_runtime_import_root` and `configure_runtime_pythonpath` helper logic to `desktop-tauri/src-tauri/src/main.rs` and apply it when constructing the command used to run the model bridge so the spawned Python process receives a correct `PYTHONPATH`. 
- Add a focused unit test `tests/unit/test_desktop_path_bootstrap.py` that simulates the `bin/python/model_bridge.py` + `bin/resources/utils` layout and an existing dev-tree layout to prevent regressions for the real desktop launch path. 

### Testing

- Ran `python -m pytest tests/unit/test_desktop_model_bridge.py tests/unit/test_desktop_path_bootstrap.py` and the targeted unit tests passed. 
- Ran `python -m pytest tests/unit/test_desktop_compute_node_bridge.py` and existing compute-node bridge unit tests passed. 
- Ran `npm --prefix desktop-tauri run build` which succeeded; `npm --prefix desktop-tauri run tauri build -- --debug --no-bundle` was attempted but failed in this environment due to missing system `glib-2.0` pkg-config/dev libs. 
- Ran `cargo test --manifest-path desktop-tauri/src-tauri/Cargo.toml` which could not complete in this environment for the same missing system dependency (`glib-2.0`).

------
[Codex Task](https://chatgpt.com/codex/cloud/tasks/task_e_69dc2f0e5b30832f899cf715f8bda3f4)

### DIFF
--- a/desktop-tauri/src-tauri/python/path_bootstrap.py
+++ b/desktop-tauri/src-tauri/python/path_bootstrap.py
@@ -10,10 +10,12 @@ def ensure_runtime_import_paths(script_file: str) -> None:
     """Add likely import roots for development and packaged desktop layouts."""
 
     script_path = Path(script_file).resolve()
-    resources_root = script_path.parent.parent
+    script_root = script_path.parent.parent
     candidates = [
-        resources_root,  # bundled resources root in packaged apps
-        resources_root / "_up_",  # tauri ".." resources are rewritten under _up_
+        script_root,  # bundled resources root in packaged apps
+        script_root / "resources",  # no-bundle/debug layout when script is under <exe>/python
+        script_root / "Resources",  # macOS-style resources casing
+        script_root / "_up_",  # tauri ".." resources are rewritten under _up_
         script_path.parent.parent.parent,
     ]
 
--- a/desktop-tauri/src-tauri/src/main.rs
+++ b/desktop-tauri/src-tauri/src/main.rs
@@ -108,13 +108,15 @@ fn repo_runtime_import_root(manifest_dir: &Path) -> Option<String> {
 }
 
 fn configure_runtime_pythonpath(command: &mut std::process::Command) {
+    // NOTE: CARGO_MANIFEST_DIR is compile-time and primarily helps local/dev launches.
+    // Packaged end-user launches rely on python/path_bootstrap.py for runtime import roots.
     let manifest_dir = Path::new(env!("CARGO_MANIFEST_DIR"));
     if let Some(import_root) = repo_runtime_import_root(manifest_dir) {
         match std::env::var("PYTHONPATH") {
             Ok(existing) if !existing.trim().is_empty() => {
-                if let Ok(joined) =
-                    std::env::join_paths([Path::new(&import_root), Path::new(&existing)])
-                {
+                let mut components = vec![PathBuf::from(&import_root)];
+                components.extend(std::env::split_paths(&existing));
+                if let Ok(joined) = std::env::join_paths(components) {
                     command.env("PYTHONPATH", joined);
                 } else {
                     command.env("PYTHONPATH", import_root);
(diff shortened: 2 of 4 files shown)


futuroptimist:
@codex please address the Codex, Copilot, and Greptile comments

futuroptimist:
@codex
Scope Lock recap: Only touch `desktop-tauri/src-tauri/src/main.rs`, `desktop-tauri/src-tauri/python/model_bridge.py`, `desktop-tauri/src-tauri/python/path_bootstrap.py`, and minimal targeted tests under `tests/unit/`; do not touch compute-node/sidecar behavior or unrelated desktop/runtime files; keep diff small; run `cargo test --manifest-path desktop-tauri/src-tauri/Cargo.toml`, `python -m pytest tests/unit/test_desktop_compute_node_bridge.py`, any new targeted pytest file(s), `npm --prefix desktop-tauri run build`, and `npm --prefix desktop-tauri run tauri build -- --debug --no-bundle`. If `git diff --stat` shows > 8 files or touches do-not-touch areas, stop and split.

This is close, but it is not 100% yet because the new coverage still does not directly exercise the `model_bridge.py` import/runtime path in a packaged-like subprocess with `PYTHONPATH` unset. The prompt explicitly called out that green compute-node e2e missed the real release-path failure, so we need one regression test that would have failed before and proves the exact model-bridge path now works.

Please do the following:
1. Add a focused Python regression test that launches `desktop-tauri/src-tauri/python/model_bridge.py` in a subprocess from a realistic packaged-like layout, with `PYTHONPATH` explicitly removed from the environment, and verifies it can import shared runtime modules successfully instead of returning the `Missing Python dependency for model downloads (No module named 'utils')` error.
   - Prefer adding this to `tests/unit/test_desktop_model_bridge.py` if that file already has nearby coverage.
   - Simulate the real desktop layout as closely as possible using temp dirs, for example a script under an `.../python/` directory with shared modules under sibling `resources/` (and, if needed, the packaged rewrite variant your fix is intended to support).
   - Assert the user-visible contract directly: successful JSON payload from `model_bridge.py inspect`, not just `sys.path` contents.

2. Keep the current `main.rs` launch-path fix intact.
   - Do not remove the `configure_runtime_pythonpath` behavior from `desktop-tauri/src-tauri/src/main.rs`.
   - Do not broaden bootstrap candidate logic further unless the subprocess regression proves an actual remaining gap.

3. Sanity-check that the test would have failed before this PR.
   - In the test docstring or nearby comment, briefly explain which pre-fix layout would have failed and why.
   - Use existing implementation patterns; do not introduce brittle cwd-dependent behavior.

Helpful searches:
- `rg -n "Missing Python dependency for model downloads|model_bridge.py|configure_runtime_pythonpath|path_bootstrap" desktop-tauri tests`
- `rg -n "inspect_model|run_model_bridge|test_desktop_model_bridge" .`

Verification:
- `python -m pytest tests/unit/test_desktop_model_bridge.py`
- `python -m pytest tests/unit/test_desktop_compute_node_bridge.py`
- `cargo test --manifest-path desktop-tauri/src-tauri/Cargo.toml`
- `npm --prefix desktop-tauri run build`
- `npm --prefix desktop-tauri run tauri build -- --debug --no-bundle`

Keep the follow-up narrowly scoped to the missing direct regression coverage for the model-bridge launch path. No unrelated refactors.

PR #703, opened 2026-04-13
🐛 Fix bundled desktop bridges failing to import utils in installed releases

futuroptimist:
### Motivation
- Packaged/installed desktop builds could still raise `No module named 'utils'` because bundled `../../utils` may be rewritten into nested `_up_` paths by Tauri and previously-used heuristics missed those layouts.
- Both Python bridges already call `ensure_runtime_import_paths(__file__)`, so the remaining issue was that the runtime import root was not being explicitly conveyed for packaged launches.
- Provide a small, explicit runtime contract from Rust to Python so the packaged layout is resolved deterministically at launch.

### Description
- Resolve the effective bundled resource root from the resolved bridge script path in Rust and export explicit env vars used by Python: `TOKEN_PLACE_RESOURCE_DIR` and `TOKEN_PLACE_PYTHON_IMPORT_ROOT`, and set `PYTHONPATH` when an import root is found; applied for model bridge, compute-node bridge and sidecar launcher (`main.rs`, `compute_node.rs`, `sidecar.rs`).
- Extend the Python bootstrap `ensure_runtime_import_paths` to prefer the explicit env vars first and then fall back to heuristic discovery, and add support for multi-level `_up_` candidates that Tauri uses when rewriting `..` segments (`desktop-tauri/src-tauri/python/path_bootstrap.py`).
- Add focused regression unit tests that simulate packaged layouts:
  - explicit-import-root precedence and nested `_up_` resolution in `tests/unit/test_desktop_path_bootstrap.py`, and
  - a compute-node subprocess regression that runs the bridge from a packaged-like nested `_up_/_up_` layout without `PYTHONPATH` in `tests/unit/test_desktop_compute_node_bridge.py`.
- Changes limited to the desktop runtime/bridge files and tests as requested.

### Testing
- Ran `python -m pytest tests/unit/test_desktop_compute_node_bridge.py tests/unit/test_desktop_model_bridge.py tests/unit/test_desktop_path_bootstrap.py` and observed all unit tests pass: `22 passed`.
- Ran `python -m pytest tests/unit/test_desktop_compute_node_bridge.py` and observed `10 passed` for the compute-node suite.
- Built the frontend with `npm --prefix desktop-tauri run build` which succeeded.
- Attempted `cargo test --manifest-path desktop-tauri/src-tauri/Cargo.toml` and `npm --prefix desktop-tauri run tauri build -- --debug --no-bundle` in this container but they fail due to missing system-level dependency `glib-2.0`/pkg-config in the execution environment (these failures are environmental and not related to the change).
- `pre-commit run --all-files` was not available in the container environment and was not run here.

Commands used during verification: `python -m pytest ...`, `npm --prefix desktop-tauri run build`, `npm --prefix desktop-tauri run tauri build -- --debug --no-bundle`, and `cargo test --manifest-path desktop-tauri/src-tauri/Cargo.toml` (env-dependent).

------
[Codex Task](https://chatgpt.com/codex/cloud/tasks/task_e_69dc761c0698832fb93156c175840249)

### DIFF
--- a/desktop-tauri/src-tauri/python/path_bootstrap.py
+++ b/desktop-tauri/src-tauri/python/path_bootstrap.py
@@ -2,8 +2,23 @@
 
 from __future__ import annotations
 
+import os
 import sys
 from pathlib import Path
+
+
+def _has_runtime_modules(path: Path) -> bool:
+    return (path / "utils").is_dir() or (path / "config.py").is_file()
+
+
+def _resource_candidates(resource_dir: Path) -> list[Path]:
+    # Tauri rewrites ".." path segments into nested "_up_" directories.
+    candidates = [resource_dir]
+    nested = resource_dir
+    for _ in range(4):
+        nested = nested / "_up_"
+        candidates.append(nested)
+    return candidates
 
 
 def ensure_runtime_import_paths(script_file: str) -> None:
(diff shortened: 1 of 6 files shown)


futuroptimist:

PR #724, opened 2026-04-15
🛠️ fix: auto-repair desktop CUDA runtime on Windows

futuroptimist:
### Motivation
- Desktop Auto mode often stayed CPU-only on Windows because the sidecar runtime path used a probe-only default instead of repairing the runtime in the interpreter the sidecars actually run with. 
- The goal is to ensure GPU-capable `llama-cpp-python` is repaired/installed in the same `sys.executable` interpreter when CUDA prerequisites exist so Auto/GPU modes use CUDA by default and fall back cleanly when they cannot.

### Description
- Make Windows GPU-preferring launches attempt a one-time in-place repair using the canonical README recipe by running `sys.executable -m pip install --force-reinstall --verbose` with `CMAKE_ARGS=-DGGML_CUDA=on` and `FORCE_CMAKE=1` as the primary repair plan, while preserving existing fallback plans; changes live in `desktop-tauri/src-tauri/python/desktop_runtime_setup.py`.
- Probe the real sidecar interpreter before and after install via a subprocess probe (`_probe_llama_runtime_from_subprocess`) to verify `backend_available`/`gpu_offload_supported` in the exact interpreter used by sidecars and only return a restart-required action when CUDA is truly active.
- Re-exec the sidecar/bridge process automatically after a successful repair so the newly installed runtime is immediately used (implemented in `inference_sidecar.py` and `compute_node_bridge.py`), and add concise `python=<executable>` observability to `desktop.runtime_setup` logs.
- Improve runtime diagnostics visibility by reporting interpreter/prefix/module-path info from `detect_llama_runtime_capabilities()` and adding a manual helper `desktop-tauri/src-tauri/python/verify_llama_runtime.py` plus targeted unit tests under `tests/unit/` to cover repair decisioning, same-interpreter pip usage, disable opt-out behavior, failure fallback, and re-exec behavior.

### Testing
- Ran static/compile checks with `python -m py_compile <files>` on modified Python modules and the new verify helper, which succeeded. 
- Ran the focused unit test suites with `pytest tests/unit/test_desktop_runtime_setup.py tests/unit/test_desktop_inference_sidecar.py tests/unit/test_desktop_compute_node_bridge.py -q` and all tests passed (`44 passed`).
- Executed the manual verification helper with `python desktop-tauri/src-tauri/python/verify_llama_runtime.py auto` to emit authoritative runtime diagnostics (printed `sys.executable`, `llama_cpp.__file__`, backend markers, offload support, and ModelManager compute diagnostics). 
- Ran `pre-commit run --all-files` which executed repository checks and reported unrelated existing repo issues (e.g., `codespell` noise in `desktop-tauri/package-lock.json` and other environment/dependency work items), so pre-commit hooks flagged unrelated failures independent of these changes.

------
[Codex Task](https://chatgpt.com/codex/cloud/tasks/task_e_69df2234e2b8832fa7a3a6202b66813c)

### DIFF
--- a/desktop-tauri/README.md
+++ b/desktop-tauri/README.md
@@ -38,20 +38,38 @@ npm ci
 npm run tauri dev
 ```
 
-During normal startup, desktop sidecars run a **probe-only** runtime check in
+During normal startup, desktop sidecars run a runtime check in
 `auto`/`gpu`/`hybrid` modes and emit:
 
 - `desktop.runtime_setup ...` during sidecar start (backend selected + fallback reason)
 - `compute_runtime ...` after `Llama(...)` init (backend actually used, offloaded
   layers, KV cache placement, and fallback reason)
 
-Normal inference requests do **not** mutate Python packages. For local desktop
-development, explicitly run one bootstrap start with
-`TOKEN_PLACE_DESKTOP_ENABLE_RUNTIME_BOOTSTRAP=1` to allow a one-time
-`llama-cpp-python` install/repair attempt, then restart sidecars/app.
+On Windows desktop builds, `auto`/`gpu`/`hybrid` now attempt a one-time
+`llama-cpp-python` CUDA repair automatically in the same interpreter used by
+the sidecar (`sys.executable`) when the active runtime is CPU-only. The repair
+uses the canonical README recipe:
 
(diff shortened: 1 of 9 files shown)


futuroptimist:

PR #738, opened 2026-04-17
Fix packaged desktop runtime fallback when requirements.txt is absent

futuroptimist:
### Motivation

- Packaged Windows desktop runtimes could crash startup with `FileNotFoundError` when `ensure_desktop_llama_runtime()` attempted to read a repo-root `requirements.txt` under a runtime-extracted `%AppData%` layout, causing the compute-node bridge to exit before emitting startup events.
- The source-repair path called `llama_cpp_requirement_spec()` unguarded which raised early when the file was missing; the goal is to make packaged layouts degrade to a safe, deterministic install plan while preserving pinned installs for normal dev/repo-root flows.

### Description

- Add `_llama_cpp_package_spec(requirements_path)` which returns the pinned `llama-cpp-python==...` when present and falls back to the unpinned string `llama-cpp-python` when the requirements file is missing, unreadable, or unpinned.
- Use `_llama_cpp_package_spec()` from `_windows_cuda_source_repair()` so source-repair no longer raises on missing `requirements.txt` in packaged layouts.
- Detect `requirements.txt` availability and, when absent, skip invoking `llama_cpp_install_plan_fallbacks(...)` and instead use `_fallback_unpinned_plans(...)`, and broaden the fallback exception handling to include `OSError`.
- Add unit tests covering the pinned-vs-fallback package-spec helper and a Windows-packaged-layout regression test asserting that missing `requirements.txt` no longer causes startup crashes and proceeds to a CPU fallback plan.

### Testing

- Ran `pytest tests/unit/test_desktop_runtime_setup.py tests/unit/test_desktop_compute_node_bridge.py` and all tests passed (`39 passed`).
- Ran the desktop Rust tests with `cargo test --manifest-path desktop-tauri/src-tauri/Cargo.toml compute_node`, which failed in this container due to a missing system `glib-2.0` / `pkg-config` dependency (environment issue), not the Python change.

------
[Codex Task](https://chatgpt.com/codex/cloud/tasks/task_e_69e1cf3da634832f89ca4ad00b09df44)

### DIFF
--- a/desktop-tauri/src-tauri/python/desktop_runtime_setup.py
+++ b/desktop-tauri/src-tauri/python/desktop_runtime_setup.py
@@ -159,7 +159,7 @@ def _windows_cuda_source_repair(requirements_path: Path) -> tuple[bool, str]:
     env = os.environ.copy()
     env["CMAKE_ARGS"] = "-DGGML_CUDA=on"
     env["FORCE_CMAKE"] = "1"
-    package_spec = llama_cpp_requirement_spec(requirements_path)
+    package_spec = _llama_cpp_package_spec(requirements_path)
     cmd = [
         sys.executable,
         "-m",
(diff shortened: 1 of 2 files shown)


futuroptimist:

PR #741, opened 2026-04-17
Fix desktop Windows GPU bootstrap to avoid premature CPU fallback

futuroptimist:
### Motivation
- Root cause 1: the Windows desktop install fallback chain skipped a CUDA source-build attempt and could drop straight to CPU when CUDA wheels were unavailable for the active Python ABI. 
- Root cause 2: runtime probe was too narrow and missed newer/alternate `llama-cpp-python` backend markers (or nested extension module markers), causing false CPU-only conclusions. 
- The observed symptom was the desktop operator/sidecar logging `installed_cpu_fallback` / `cpu_fallback` even when CUDA could be available per README guidance. 

### Description
- Added a Windows CUDA source-build fallback plan (CMAKE_ARGS `-DGGML_CUDA=on`, `FORCE_CMAKE=1`, `--no-binary llama-cpp-python`) to the packaging planner and the runtime unpinned fallback list so desktop bootstrap will try a source build before falling back to CPU. 
- Hardened `detect_llama_runtime_capabilities()` to accept multiple marker name variants and to probe both top-level and nested `llama_cpp` modules (markers checked include `GGML_USE_CUDA`, `GGML_CUDA`, `GGML_USE_CUBLAS`, `GGML_USE_METAL`, `GGML_METAL`). 
- Added unit tests covering the new Windows fallback ordering and source-build success path, packaging plan expectations (pip args/env), marker-variant runtime detection, and a higher-level desktop bridge test that asserts GPU-capable status is emitted when runtime + compute diagnostics report CUDA. 
- Documented the desktop/Tauri bootstrap fallback chain in `README.md` so the runtime behavior matches the Windows GPU guidance (CUDA wheel -> unpinned CUDA wheel -> CUDA source build -> CPU fallback). 

### Testing
- Ran targeted unit tests: `pytest tests/unit/test_desktop_runtime_setup.py tests/unit/test_desktop_compute_node_bridge.py tests/unit/test_desktop_gpu_packaging.py tests/unit/test_model_manager.py`, all unit tests passed (103 passed locally in the CI-like environment). 
- Confirmed the new unit that simulates wheel-unavailable then source-build success exercises the new path and asserts `installed_cuda_reexec` is returned. 
- Tried `cargo test --manifest-path desktop-tauri/src-tauri/Cargo.toml compute_node`, which failed in this environment due to missing system `glib-2.0`/pkg-config system libs (environmental/system dependency, unrelated to the changes). 
- Ran `pre-commit run --all-files`; it executed but exposed existing repository-level hooks (codespell and vulture findings) unrelated to this change (pre-existing failures), so pre-commit did not fully pass in this environment.

------
[Codex Task](https://chatgpt.com/codex/cloud/tasks/task_e_69e1d86aadec832f86409f9271c9a37e)

### DIFF
--- a/README.md
+++ b/README.md
@@ -474,6 +474,10 @@ The next steps need to be executed in the same virtual environment you set up ab
 
 This will replace the llama-cpp-python you installed via `pip install -r requirements.txt` and will instruct it to use [cuBLAS](https://docs.nvidia.com/cuda/cublas/index.html).
 
+For the desktop Tauri compute-node path, runtime bootstrap follows the same contract:
+it first tries a CUDA wheel, then an unpinned CUDA wheel, then a CUDA source build
+with `CMAKE_ARGS=-DGGML_CUDA=on` and `FORCE_CMAKE=1`, and only then falls back to CPU.
+
 **if you're using Command Prompt**
 
 ```
--- a/desktop-tauri/src-tauri/python/desktop_gpu_packaging.py
+++ b/desktop-tauri/src-tauri/python/desktop_gpu_packaging.py
@@ -129,6 +129,22 @@ def llama_cpp_install_plan_fallbacks(
             )
         )
 
+        # If no compatible CUDA wheels exist (common with new Python ABIs),
+        # try a source build with CUDA explicitly enabled before giving up.
+        plans.append(
+            LlamaCppInstallPlan(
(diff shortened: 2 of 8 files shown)


futuroptimist:

PR #751, opened 2026-04-18
Fix desktop runtime root resolution to unblock sidecar & compute-node startup

futuroptimist:
### Motivation
- Packaged Tauri launches could probe the wrong import/cwd root (the code used a brittle `Path(__file__).parents[3]` assumption), causing the runtime probe to fail and triggering lengthy bootstrap/repair logic before the sidecar/bridge emitted their first structured `started` event.
- The shared startup path (`ensure_desktop_llama_runtime`) is called by both the inference sidecar and compute-node bridge, so a broken probe prevented both local inference and relay registration from ever reaching a usable state.

### Description
- Add `_resolve_runtime_root` which prefers `TOKEN_PLACE_PYTHON_IMPORT_ROOT` and otherwise searches up from the script for a directory containing `utils/` or `config.py`, falling back to the previous heuristic when necessary.
- Make `_probe_llama_runtime` accept a `runtime_root` and use that resolved root for the probe subprocess `cwd` and `PYTHONPATH`, and add a `_probe_runtime` wrapper for backward-compatible tests.
- Replace fixed `parents[3]` usage in `ensure_desktop_llama_runtime` with the resolved `target_root` and call the new `_probe_runtime` helper at all probe points.
- Harden runtime state persistence by catching `OSError` when reading/writing `~/.token_place_desktop_runtime_state.json` to avoid crashing before structured events are emitted.
- Add focused regression tests in `tests/unit/test_desktop_runtime_setup.py` verifying import-root preference and that the probe subprocess uses the resolved `cwd` and `PYTHONPATH`, plus small test adjustments for compatibility with mocked `subprocess.run` signatures.

### Testing
- Ran targeted unit tests: `pytest -q --noconftest tests/unit/test_desktop_runtime_setup.py tests/unit/test_desktop_compute_node_bridge.py`, all tests passed (`48 passed`).
- Ran repository checks via `pre-commit run --all-files`; pre-commit hooks run but the full project check suite failed in this container due to a missing system dependency in the environment (`cryptography`), not due to the patch logic itself.
- Verified the new tests exercise the probe cwd/PYTHONPATH behavior and the runtime-state read/write hardening, and existing compute-node bridge tests remained green.

------

### DIFF
--- a/desktop-tauri/src-tauri/python/desktop_runtime_setup.py
+++ b/desktop-tauri/src-tauri/python/desktop_runtime_setup.py
@@ -74,8 +74,26 @@ print(json.dumps(payload))
 """.strip()
 
 
-def _probe_llama_runtime() -> RuntimeProbe:
-    repo_root = Path(__file__).resolve().parents[3]
+def _resolve_runtime_root(*, repo_root: Optional[Path] = None) -> Path:
+    if repo_root is not None:
+        return repo_root.resolve()
+
+    explicit_root = os.environ.get("TOKEN_PLACE_PYTHON_IMPORT_ROOT", "").strip()
+    if explicit_root:
+        candidate = Path(explicit_root).resolve()
+        if (candidate / "utils").is_dir() or (candidate / "config.py").is_file():
+            return candidate
+
+    script_path = Path(__file__).resolve()
+    for candidate in script_path.parents:
+        if (candidate / "utils").is_dir() or (candidate / "config.py").is_file():
+            return candidate
+
+    return script_path.parents[3]
+
+
+def _probe_llama_runtime(*, runtime_root: Optional[Path] = None) -> RuntimeProbe:
+    repo_root = _resolve_runtime_root(repo_root=runtime_root)
     python_root = Path(__file__).resolve().parent
     cmd = [sys.executable, "-c", _PROBE_SNIPPET]
     env = os.environ.copy()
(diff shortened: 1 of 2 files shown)


futuroptimist:

futuroptimist:
@codex
Scope Lock recap: Only touch `desktop-tauri/src-tauri/python/desktop_runtime_setup.py`, `tests/unit/test_desktop_runtime_setup.py`, and the PR description if needed; do not touch UI/Tauri Rust/JS files, `utils/*`, or relay code; keep diff small; run `pytest -q --noconftest tests/unit/test_desktop_runtime_setup.py tests/unit/test_desktop_compute_node_bridge.py` and `pre-commit run --all-files`.

This is the best candidate, but it is not 100% yet because the regression coverage is still one level too indirect for the exact prompt-scoped contract.

Remaining work:
1. Add one focused regression test proving that `ensure_desktop_llama_runtime(..., repo_root=<custom root>)` uses that exact root for the initial probe and the post-repair reprobe path, rather than only proving `_probe_llama_runtime()` honors the env var. The prompt explicitly cares about deterministic startup in the shared bootstrap path, so cover the public/shared entrypoint, not just the helper.
2. Tighten `_resolve_runtime_root()` / associated tests so explicit import-root preference is validated against a token.place-like layout (`utils/` or `config.py`), and make the fallback behavior explicit in tests. The goal is to prevent accepting an arbitrary existing directory and silently probing the wrong place in packaged layouts.
3. Update the PR summary to be explicit about:
   - the exact root cause,
   - why both regressions were related,
   - the exact verification commands run,
   - whether local relay registration and local inference were verified manually or only via tests.
   Right now the summary is close, but it still does not fully satisfy the original deliverables language.

Helpful search/check commands:
- `rg -n "_probe_llama_runtime|_probe_runtime|ensure_desktop_llama_runtime|TOKEN_PLACE_PYTHON_IMPORT_ROOT|TOKEN_PLACE_PROBE_REPO_ROOT" desktop-tauri/src-tauri/python/desktop_runtime_setup.py tests/unit/test_desktop_runtime_setup.py`
- `pytest -q --noconftest tests/unit/test_desktop_runtime_setup.py tests/unit/test_desktop_compute_node_bridge.py`
- `pre-commit run --all-files`

Diff sanity check: If `git diff --stat` shows > 3 files or touches do-not-touch areas, stop and split.

PR #754, opened 2026-04-18
Fix desktop runtime probe root resolution for sidecar/compute-node startup

futuroptimist:
### Motivation
- The probe used by desktop sidecar/compute-node always used a hardcoded repo root (`parents[3]`), which can be incorrect in packaged desktop layouts and causes the probe/bootstrap to run from the wrong cwd and PYTHONPATH.
- When the runtime probe runs from the wrong root the bootstrap may attempt unnecessary or failing repair steps and block before emitting the first structured `started` event, leaving both inference and compute-node registration pending.
- The change makes the probe honor the import root set by the Tauri launcher so packaged launches probe the correct runtime layout and emit startup events promptly.

### Description
- Added `_resolve_runtime_root()` to `desktop-tauri/src-tauri/python/desktop_runtime_setup.py` that prefers `TOKEN_PLACE_PYTHON_IMPORT_ROOT` and falls back to packaged/layout candidates when present.
- Updated `_probe_llama_runtime()` to use the resolved runtime root for `PYTHONPATH`, `TOKEN_PLACE_PROBE_REPO_ROOT`, and the probe subprocess `cwd` so imports and probe behavior align with the desktop import root.
- Updated `ensure_desktop_llama_runtime()` to use the same resolver for its default `target_root` selection, preserving consistent bootstrap behavior.
- Added two focused unit tests in `tests/unit/test_desktop_runtime_setup.py`: `test_resolve_runtime_root_prefers_explicit_import_root_env` and `test_probe_uses_resolved_runtime_root_for_pythonpath_and_cwd` and a small `os` import required by the tests.

### Testing
- Ran unit suites: `pytest -q --noconftest tests/unit/test_desktop_runtime_setup.py tests/unit/test_desktop_compute_node_bridge.py` and observed all tests pass (`48 passed`).
- Attempted `pre-commit run --all-files` but `pre-commit` is not installed in this execution environment, so the hook was not executed here.
- Manual end-to-end verification against a real GGUF + local `relay.py` was not performed in this environment; verification was done via the updated unit tests that cover the resolved root behavior.

------

### DIFF
--- a/desktop-tauri/src-tauri/python/desktop_runtime_setup.py
+++ b/desktop-tauri/src-tauri/python/desktop_runtime_setup.py
@@ -74,19 +74,42 @@ print(json.dumps(payload))
 """.strip()
 
 
+def _resolve_runtime_root() -> Path:
+    explicit_import_root = os.environ.get("TOKEN_PLACE_PYTHON_IMPORT_ROOT", "").strip()
+    if explicit_import_root:
+        candidate = Path(explicit_import_root).resolve()
+        if candidate.exists():
+            return candidate
+
+    script_path = Path(__file__).resolve()
+    candidates = [
+        script_path.parents[3] if len(script_path.parents) > 3 else None,
+        script_path.parent.parent,
+        script_path.parent.parent / "resources",
+        script_path.parent.parent / "Resources",
+    ]
+    for candidate in candidates:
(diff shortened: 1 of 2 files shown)


futuroptimist:

PR #755, opened 2026-04-18
🐛 Fix relay signal shutdown deadlock

futuroptimist:
### Motivation
- The relay process could hang after receiving SIGINT/SIGTERM because `server.shutdown()` was invoked directly from the signal handler while the server thread was running `serve_forever()`, preventing timely termination and readiness updates.

### Description
- Replace the inline signal handler inside `serve()` to call a new helper ` _request_relay_shutdown(server, shutdown_requested, signum)` instead of calling `server.shutdown()` directly.
- Add `_shutdown_server_async(server)` to invoke `server.shutdown()` on a background daemon thread and make shutdown initiation idempotent. 
- Ensure the global `DRAINING` flag is set immediately when a shutdown signal is requested so `/healthz` reflects draining right away. 
- Add regression tests in `tests/unit/test_relay_shutdown_signal.py` that assert async shutdown is performed on a separate thread and that duplicate signals are idempotent.

### Testing
- `python -m py_compile relay.py tests/unit/test_relay_shutdown_signal.py` succeeded.
- Attempting `pytest -q tests/unit/test_relay_shutdown_signal.py` in this environment failed due to a missing test dependency (`ModuleNotFoundError: No module named 'requests'` from `tests/conftest.py`).
- `pre-commit run --all-files` could not be executed in this environment because the `pre-commit` command is not installed.

------

### DIFF
--- a/relay.py
+++ b/relay.py
@@ -889,9 +889,7 @@ def serve(host: str, port: int) -> None:
     shutdown_requested = threading.Event()
 
     def _handle_signal(signum, _frame):
-        LOGGER.info("relay.shutdown_signal", extra={"signal": signum})
-        shutdown_requested.set()
-        server.shutdown()
+        _request_relay_shutdown(server, shutdown_requested, signum)
 
     signal.signal(signal.SIGTERM, _handle_signal)
     signal.signal(signal.SIGINT, _handle_signal)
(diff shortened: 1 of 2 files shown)


futuroptimist:

PR #757, opened 2026-04-18
🐛 Fix relay SIGINT/SIGTERM shutdown deadlock

futuroptimist:
### Motivation
- Calling `BaseServer.shutdown()` directly from a signal handler can deadlock when the serving loop (`serve_forever()`) runs on the same thread, preventing `Ctrl+C`/`SIGTERM` from terminating the process promptly.
- Ensure the relay process enters draining and terminates quickly after receiving shutdown signals while preserving readiness/liveness behavior.
- Add regression coverage to prevent reintroduction of the deadlock in future changes.

### Description
- Change `serve()` signal handler to run `server.shutdown()` on a background daemon thread named `relay-shutdown` instead of calling it inline in the handler to avoid deadlocking the main serve thread (`relay.py`).
- Add a regression test `test_serve_signal_handler_runs_shutdown_off_main_thread` that monkeypatches `make_server`/`signal`/`Thread` to assert shutdown is invoked from a background thread and that `serve()` triggers the handler when `serve_forever()` receives a signal (`tests/test_relay.py`).
- Add small test imports and module-level `relay` reference required by the new regression test (`tests/test_relay.py`).

### Testing
- Ran `pytest -q tests/test_relay.py -k "serve_signal_handler_runs_shutdown_off_main_thread or healthz_returns_draining_when_shutdown_flag_set"`, which attempted to run the new regression test but failed to load test bootstrap due to a missing runtime dependency (`ModuleNotFoundError: No module named 'requests'`).
- Attempted `pre-commit run --all-files` which could not be executed in this environment because the `pre-commit` command is not installed.
- Secret-scan hook (`./scripts/scan-secrets.py`) is not present in this environment so that check was skipped.

------

### DIFF
--- a/relay.py
+++ b/relay.py
@@ -891,7 +891,13 @@ def serve(host: str, port: int) -> None:
     def _handle_signal(signum, _frame):
         LOGGER.info("relay.shutdown_signal", extra={"signal": signum})
         shutdown_requested.set()
-        server.shutdown()
+        # BaseServer.shutdown() must be invoked from a different thread than
+        # serve_forever() to avoid deadlocking while handling SIGINT/SIGTERM.
+        threading.Thread(
+            target=server.shutdown,
+            name="relay-shutdown",
+            daemon=True,
+        ).start()
 
     signal.signal(signal.SIGTERM, _handle_signal)
     signal.signal(signal.SIGINT, _handle_signal)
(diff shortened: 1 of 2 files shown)


futuroptimist:

PR #758, opened 2026-04-18
🐛 Fix relay SIGINT shutdown deadlock

futuroptimist:
### Motivation
- The relay process could log receipt of `SIGINT`/`SIGTERM` (e.g. when pressing Ctrl+C) but then hang because `server.shutdown()` was invoked inline from the signal handler and could block waiting on the serving loop. 
- Make termination responsive and observable while preserving draining state for readiness probes and logs.

### Description
- Move shutdown request off the signal handler stack by adding a `shutdown_dispatched` `threading.Event` and an `_dispatch_server_shutdown()` helper that spawns a daemon thread to call `server.shutdown()`.
- Update the runtime signal handler to set `DRAINING`, mark `shutdown_requested`, and call `_dispatch_server_shutdown()` instead of calling `server.shutdown()` inline.
- Emit `dispatched` alongside `requested` in the final `relay.shutdown` log for clearer telemetry.
- Add a regression test `tests/unit/test_relay_signal_shutdown.py` that monkeypatches `make_server` and `signal` to verify the handler returns quickly while shutdown executes asynchronously.

### Testing
- Ran `python -m py_compile relay.py tests/unit/test_relay_signal_shutdown.py`, which succeeded.
- Attempted `pytest -q tests/unit/test_relay_signal_shutdown.py`, which failed in this environment due to a missing dependency (`tests/conftest.py` imports `requests`), not due to the new test itself.
- `pre-commit run --all-files` was attempted but is not available in this environment (`pre-commit: command not found`).

------

### DIFF
--- a/relay.py
+++ b/relay.py
@@ -887,11 +887,25 @@ def serve(host: str, port: int) -> None:
     ctx.push()
 
     shutdown_requested = threading.Event()
+    shutdown_dispatched = threading.Event()
+
+    def _dispatch_server_shutdown() -> None:
+        """Request server shutdown without blocking the signal handler thread."""
+
+        if shutdown_dispatched.is_set():
+            return
+        shutdown_dispatched.set()
+        threading.Thread(
+            target=server.shutdown,
+            name="relay-server-shutdown",
+            daemon=True,
+        ).start()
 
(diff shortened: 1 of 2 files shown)


futuroptimist:

PR #763, opened 2026-04-19
Gate desktop runtime auto-repair behind opt-in to avoid startup stalls

futuroptimist:
### Motivation
- The desktop bridge and sidecar call `ensure_desktop_llama_runtime(...)` before emitting any startup events, which allowed heavyweight Windows GPU repair/install flows to run during normal startup and stall the process indefinitely. 
- This early blocking runtime bootstrap is the real cause of the UI/operator hang (bridge never reached `started`/poll loop); PR 760 fixed late relay heartbeat handling but did not address this earlier bootstrap stall. 
- We need a minimal, deterministic change so ordinary desktop launches are fast and only explicit repair flows perform long-running installs/rebuilds.

### Description
- Add an explicit opt-in environment gate `TOKEN_PLACE_DESKTOP_ENABLE_RUNTIME_BOOTSTRAP` and define `ENABLE_BOOTSTRAP_ENV` in `desktop_runtime_setup.py`. 
- On Windows, if the opt-in is not set, `ensure_desktop_llama_runtime` returns probe-only immediately with a clear `fallback_reason` and `runtime_action='probe_only'`, avoiding implicit pip/source-build/reexec work during normal startup. 
- Preserve the existing repair/install + re-exec behavior when `TOKEN_PLACE_DESKTOP_ENABLE_RUNTIME_BOOTSTRAP=1` so dedicated repair flows continue to work. 
- Update unit tests in `tests/unit/test_desktop_runtime_setup.py` to opt-in where necessary and add `test_windows_runtime_bootstrap_defaults_to_probe_only_without_opt_in` to cover the new default fast-path.

### Testing
- Ran unit tests: `pytest -q --noconftest tests/unit/test_desktop_runtime_setup.py tests/unit/test_desktop_compute_node_bridge.py`, result: 59 passed. 
- Verified Python modules compile: `python -m py_compile desktop-tauri/src-tauri/python/desktop_runtime_setup.py desktop-tauri/src-tauri/python/compute_node_bridge.py desktop-tauri/src-tauri/python/inference_sidecar.py`, result: success. 
- Attempted Rust unit tests: `cd desktop-tauri/src-tauri && cargo test` failed in CI environment due to missing system `glib-2.0` / pkg-config dependency (environment limitation). 
- Attempted frontend tests: `cd desktop-tauri && npm test -- --runInBand` failed in this environment because the vitest runner does not accept `--runInBand` (flag incompatibility).

------

### DIFF
--- a/desktop-tauri/scripts/verify_desktop_runtime.py
+++ b/desktop-tauri/scripts/verify_desktop_runtime.py
@@ -27,7 +27,10 @@ def main() -> int:
     repo_llama_shim = str((repo_root / 'llama_cpp.py').resolve())
 
     from utils.llm.model_manager import detect_llama_runtime_capabilities
-    from desktop_runtime_setup import ensure_desktop_llama_runtime
+    from desktop_runtime_setup import ENABLE_BOOTSTRAP_ENV, ensure_desktop_llama_runtime
+
+    # This manual verification flow is an explicit runtime-repair entrypoint.
+    os.environ.setdefault(ENABLE_BOOTSTRAP_ENV, '1')
 
     runtime_setup = ensure_desktop_llama_runtime(args.mode, repo_root=repo_root)
     runtime = detect_llama_runtime_capabilities()
--- a/desktop-tauri/scripts/windows_nvidia_gpu_smoke_test.py
+++ b/desktop-tauri/scripts/windows_nvidia_gpu_smoke_test.py
@@ -35,10 +35,11 @@ def _offloaded_layer_count(value: Any) -> int:
 
 
 def _load_compute_runtime_diagnostics(model_path: str, mode: str) -> dict[str, Any]:
-    from desktop_runtime_setup import ensure_desktop_llama_runtime
+    from desktop_runtime_setup import ENABLE_BOOTSTRAP_ENV, ensure_desktop_llama_runtime
     from utils.compute_node_runtime import apply_compute_mode, compute_mode_diagnostics
     from utils.llm.model_manager import get_model_manager
 
+    os.environ.setdefault(ENABLE_BOOTSTRAP_ENV, '1')
     runtime_setup = ensure_desktop_llama_runtime(mode)
 
     manager = get_model_manager()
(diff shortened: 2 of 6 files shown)


futuroptimist:
@codex please address the Codex comments

futuroptimist:
@codex
Scope Lock recap: No scope lock provided. Only touch `desktop-tauri/src-tauri/python/desktop_runtime_setup.py`, targeted desktop startup tests (prefer `tests/unit/test_desktop_runtime_setup.py` and only the minimal additional desktop-sidecar/bridge test if needed), and no unrelated docs/UI/relay transport changes; keep diff small; run these commands.

This is the right direction, but it is not 100% yet.

Remaining work:
1. Add one focused regression proving the fix actually covers the second user-facing symptom (local inference startup), not just the shared bootstrap helper.
   - Prefer adding/updating a targeted test in `tests/unit/test_desktop_inference_sidecar.py` or the smallest existing desktop-sidecar test file.
   - Assert that when Windows GPU runtime is unavailable and `TOKEN_PLACE_DESKTOP_ENABLE_RUNTIME_BOOTSTRAP` is not set, startup takes the probe-only path without invoking repair/install work and still reaches the first structured startup boundary instead of hanging.
   - Keep it deterministic: mock `_probe_llama_runtime` / repair hooks rather than relying on real CUDA/runtime state.

2. Add one focused bridge-side assertion if it is currently missing from prompt-scoped coverage.
   - Search:
     - `rg -n "probe_only|ENABLE_BOOTSTRAP_ENV|desktop.runtime_setup|started" tests/unit`
     - `rg -n "ensure_desktop_llama_runtime|runtime_action" tests/unit/test_desktop_compute_node_bridge.py tests/unit/test_desktop_inference_sidecar.py`
   - If there is no direct startup-path test for the bridge/operator flow under the new default, add the smallest one that proves the bridge does not attempt repair/install work during normal startup.

3. Keep the change narrowly scoped to the bootstrap default behavior.
   - Do not add README/docs churn here.
   - Do not touch relay/client/UI files.
   - Do not refactor unrelated test helpers unless required for the new regression case.

4. In the PR summary, be explicit that:
   - PR 760 did not fix the issue because it only affected late relay registration handling after startup, while this change fixes the earlier shared runtime bootstrap stall before first events.
   - This PR changes default startup behavior to probe-only unless `TOKEN_PLACE_DESKTOP_ENABLE_RUNTIME_BOOTSTRAP=1`.

Verification:
- `pytest -q --noconftest tests/unit/test_desktop_runtime_setup.py tests/unit/test_desktop_compute_node_bridge.py`
- Add the inference-sidecar test file if you touch it, e.g.:
  - `pytest -q --noconftest tests/unit/test_desktop_runtime_setup.py tests/unit/test_desktop_compute_node_bridge.py tests/unit/test_desktop_inference_sidecar.py`
- `python -m py_compile desktop-tauri/src-tauri/python/desktop_runtime_setup.py desktop-tauri/src-tauri/python/compute_node_bridge.py desktop-tauri/src-tauri/python/inference_sidecar.py`
- If the repo env supports it:
  - `cd desktop-tauri/src-tauri && cargo test`
  - `cd desktop-tauri && npm test -- --runInBand`

Diff sanity check: If `git diff --stat` shows > 8 files or touches do-not-touch areas, stop and split.

PR #764, opened 2026-04-19
Make desktop runtime bootstrap opt-in to avoid startup hangs

futuroptimist:
### Motivation
- The compute-node bridge and inference sidecar were stalling before emitting their first structured events because `ensure_desktop_llama_runtime(...)` ran heavyweight install/repair work during normal startup when a GPU-capable runtime was not present. 
- This early bootstrap stall left the UI in indefinite `pending` state and blocked both operator registration and local inference startup. 
- Prior fix in PR 760 addressed late relay/heartbeat handling but did not touch the early shared runtime bootstrap path that is the true root cause. 

### Description
- Introduced an opt-in env `TOKEN_PLACE_DESKTOP_ENABLE_RUNTIME_BOOTSTRAP` and made install/repair bootstrap probe-only by default to avoid blocking startup. 
- Preserved the existing `TOKEN_PLACE_DESKTOP_DISABLE_RUNTIME_BOOTSTRAP` behavior and added a clear fallback message that instructs how to opt into bootstrap. 
- Updated `desktop-tauri/src-tauri/python/desktop_runtime_setup.py` to return `runtime_action: probe_only` unless the new enable env is set when Windows auto-repair would otherwise run. 
- Updated unit tests in `tests/unit/test_desktop_runtime_setup.py` to explicitly set `ENABLE_BOOTSTRAP_ENV` for tests that exercise install/repair flows and added a regression test asserting probe-only default; updated docs in `desktop-tauri/README.md`. 

### Testing
- Ran `pytest -q --noconftest tests/unit/test_desktop_runtime_setup.py tests/unit/test_desktop_compute_node_bridge.py` and all tests passed (`59 passed`). 
- Ran `python -m py_compile desktop-tauri/src-tauri/python/desktop_runtime_setup.py desktop-tauri/src-tauri/python/compute_node_bridge.py desktop-tauri/src-tauri/python/inference_sidecar.py` and compilation succeeded. 
- Ran `cd desktop-tauri && npm test` and the frontend test suite passed; attempting `npm test -- --runInBand` failed due to Vitest option mismatch (not a regression). 
- Ran `cd desktop-tauri/src-tauri && cargo test` but it could not complete in this environment due to missing system `glib-2.0`/`pkg-config` (environment limitation, unrelated to the change).

------

### DIFF
--- a/desktop-tauri/README.md
+++ b/desktop-tauri/README.md
@@ -38,17 +38,17 @@ npm ci
 npm run tauri dev
 ```
 
-During normal startup, desktop sidecars probe the active sidecar interpreter and, on
-Windows in `auto`/`gpu`/`hybrid` modes, automatically run a one-time CUDA runtime
-repair when the runtime is CPU-only. They emit:
+During normal startup, desktop sidecars probe the active sidecar interpreter and
+emit:
 
 - `desktop.runtime_setup ...` during sidecar start (backend selected + fallback reason)
 - `compute_runtime ...` after `Llama(...)` init (backend actually used, offloaded
   layers, KV cache placement, and fallback reason)
 
-Set `TOKEN_PLACE_DESKTOP_DISABLE_RUNTIME_BOOTSTRAP=1` to explicitly disable the
-Windows auto-repair path and keep startup in probe-only mode (useful for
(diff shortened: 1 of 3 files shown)


futuroptimist:

PR #765, opened 2026-04-19
🐛 Gate desktop runtime bootstrap behind opt-in to avoid startup stalls

futuroptimist:
### Motivation
- The desktop bridge and sidecar call `ensure_desktop_llama_runtime()` before emitting first structured startup events, allowing expensive Windows repair/install work to run during ordinary startup and stall the process.  
- This caused the UI to remain at pending (no `desktop.runtime_setup` / `started` logs) while the child process hung in bootstrap.  
- PR 760 only fixed later relay heartbeat handling after the bridge reached its poll loop, so it did not address this earlier runtime-bootstrap stall.

### Description
- Add an opt-in gate `TOKEN_PLACE_DESKTOP_ENABLE_RUNTIME_BOOTSTRAP` and default to a fast probe-only path when GPU runtime is missing so normal desktop startup does not trigger repair/install actions.  
- Preserve existing `TOKEN_PLACE_DESKTOP_DISABLE_RUNTIME_BOOTSTRAP=1` behavior and keep its precedence.  
- Update unit tests in `tests/unit/test_desktop_runtime_setup.py` to explicitly set the opt-in env for tests that exercise install/repair/reexec logic and add a regression test asserting default probe-only behavior.  
- No changes to relay polling or heartbeat logic were made; the fix narrows scope to runtime bootstrap ordering and opt-in behavior in `desktop_runtime_setup.py`.

### Testing
- Ran unit tests with `pytest -q --noconftest tests/unit/test_desktop_runtime_setup.py tests/unit/test_desktop_compute_node_bridge.py` and all tests passed (`59 passed`).  
- Performed Python compile check with `python -m py_compile desktop-tauri/src-tauri/python/desktop_runtime_setup.py desktop-tauri/src-tauri/python/compute_node_bridge.py desktop-tauri/src-tauri/python/inference_sidecar.py` which completed successfully.  
- Ran `npm test` in `desktop-tauri` which passed the JS unit tests, while `npm test -- --runInBand` is incompatible with the current Vitest CLI (reported as unsupported).  
- Ran `cargo test` in `desktop-tauri/src-tauri` but it failed in this environment due to a missing system dependency (`glib-2.0` / pkg-config) unrelated to the code change.

------

### DIFF
--- a/desktop-tauri/src-tauri/python/desktop_runtime_setup.py
+++ b/desktop-tauri/src-tauri/python/desktop_runtime_setup.py
@@ -35,6 +35,7 @@ PIP_SOURCE_BUILD_TIMEOUT_SECONDS = 1800
 INSTALL_ERROR_SUMMARY_MAX_LEN = 512
 REEXEC_GUARD_ENV = "TOKEN_PLACE_DESKTOP_RUNTIME_REEXECED"
 DISABLE_BOOTSTRAP_ENV = "TOKEN_PLACE_DESKTOP_DISABLE_RUNTIME_BOOTSTRAP"
+ENABLE_BOOTSTRAP_ENV = "TOKEN_PLACE_DESKTOP_ENABLE_RUNTIME_BOOTSTRAP"
 SOURCE_REPAIR_COOLDOWN_SECONDS = 24 * 60 * 60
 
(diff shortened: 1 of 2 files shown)


futuroptimist:

PR #776, opened 2026-04-19
🐛 Fix desktop operator stop lock contention in compute-node lifecycle

futuroptimist:
### Motivation
- The operator stop path could deadlock because `start_compute_node` held `lifecycle_lock` for the entire bridge stdout/event loop lifetime and `stop_compute_node` attempted to acquire the same lock before sending the cancel message. 
- That contention prevented the cancel JSON from ever reaching the running bridge, producing a no-op/hang when the UI requested stop. 

### Description
- Move `lifecycle_lock` usage in `start_compute_node` to short critical sections only: acquire during startup/child/stdin/status initialization and re-acquire only around final teardown, then explicitly drop the lock before entering the long-running stdout loop. 
- Remove the global lock acquisition from `stop_compute_node` so a cancel JSON (`{"type":"cancel"}`) can be written immediately to the bridge stdin; mutate `child`, `stdin`, and `status` only inside short, localized critical sections protected by `lifecycle_lock`. 
- Preserve existing invariants: prevent concurrent operator sessions by checking `state.child` at startup, and ensure `state.child`, `state.stdin`, and `state.status` are updated coherently within locked sections. 
- Keep graceful-first shutdown semantics: send cancel on stdin, poll for a short window for normal exit, and fall back to `kill()` if needed, while ensuring final status ends `running=false` and `registered=false`. 
- Add two focused Rust tests that would have caught the regression: `stop_compute_node_does_not_wait_on_lifecycle_lock` and `stop_compute_node_sends_cancel_and_updates_stopped_status`. 

### Testing
- Ran Python unit suite with `pytest -q --noconftest tests/unit/test_desktop_compute_node_bridge.py` and it passed (23 tests). 
- Ran desktop JS tests with `npm --prefix desktop-tauri run test -- src/App.test.tsx` and they passed (5 tests). 
- Attempted `cargo test --manifest-path desktop-tauri/src-tauri/Cargo.toml` and `cargo test ... stop_compute_node`, but Rust build failed in this environment due to missing system `glib-2.0` dev libs (pkg-config/`glib-2.0.pc`), so the new Rust tests could not be executed here. 
- `pre-commit run --all-files` was not executed because `pre-commit` is not installed in the execution environment.

------

### DIFF
--- a/desktop-tauri/src-tauri/src/compute_node.rs
+++ b/desktop-tauri/src-tauri/src/compute_node.rs
@@ -323,9 +323,9 @@ pub async fn start_compute_node(
     state: ComputeNodeState,
     request: ComputeNodeRequest,
 ) -> anyhow::Result<()> {
-    let _lifecycle_lock = state.lifecycle_lock.lock().await;
     let manifest_dir = Path::new(env!("CARGO_MANIFEST_DIR"));
 
+    let _lifecycle_lock = state.lifecycle_lock.lock().await;
     {
         let mut child_slot = state.child.lock().await;
         if child_slot
@@ -440,6 +440,7 @@ pub async fn start_compute_node(
             last_error: None,
         };
     }
+    drop(_lifecycle_lock);
 
     let log_policy = SubprocessLogPolicy::from_env();
     let stderr_task = tokio::spawn(async move {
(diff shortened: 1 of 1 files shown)


futuroptimist:

PR #782, opened 2026-04-20
Add explicit relay unregister flow on compute-node shutdown

futuroptimist:
### Motivation
- Ensure a clean, immediate removal of stopped compute nodes from the relay so the desktop operator stop path does not rely on stale TTL eviction.
- Keep the protocol surface minimal by adding a single-purpose unregister endpoint and reusing existing conventions for server-token auth.
- Make shutdown best-effort: attempts to unregister should not block or fail operator shutdown if the relay is unreachable.

### Description
- Added a new `/unregister` relay endpoint that enforces the same `X-Relay-Server-Token` authentication as `/sink` and `/source` and removes the supplied `server_public_key` from `known_servers`.  The endpoint also returns whether a server was removed. (`relay.py`)
- Centralized per-server cleanup into `_remove_server_state(server_public_key)` and reused it from `_evict_stale_servers()` and the new `/unregister` endpoint so eviction and explicit unregister share identical cleanup behavior. (`relay.py`)
- Implemented a best-effort client-side unregister in `RelayClient.unregister_from_relay()` and call it from `RelayClient.stop()` during shutdown; the client sends the registration token header when configured and tries the active relay first, then fallbacks. (`utils/networking/relay_client.py`)
- Made `ComputeNodeRuntime.stop()` resilient by catching errors from the relay client stop/unregister path so runtime shutdown remains successful even if unregister fails. (`utils/compute_node_runtime.py`)
- Added focused unit tests covering unregister auth/behavior, client unregister/token behavior, and best-effort shutdown semantics. (`tests/unit/test_relay_registration_token.py`, `tests/unit/test_relay_client.py`, `tests/unit/test_compute_node_runtime.py`)

### Testing
- Ran `pytest -q tests/unit/test_relay_client.py` and all tests in that file passed (45 passed). 
- Ran `pytest -q tests/unit/test_desktop_compute_node_bridge.py` and all tests in that file passed (23 passed).
- Ran `pytest -q tests/unit/test_relay_registration_token.py` and `pytest -q tests/unit/test_compute_node_runtime.py` and both passed (5 passed and 21 passed respectively), verifying unregister auth, immediate diagnostics update, and runtime stop behavior.
- Ran `pre-commit run --all-files`; standard hooks completed but the integrated `run-checks` test harness reported unrelated environment-level failures in this runner (Playwright/browser artifacts and integration checks), so project hooks mostly passed while the full project check step failed in this CI-like environment and is not caused by this change.

------
[Codex Task](https://chatgpt.com/codex/cloud/tasks/task_e_69e5d6ca0414832f8be0549734f673a2)

### DIFF
--- a/relay.py
+++ b/relay.py
@@ -389,9 +389,32 @@ def _evict_stale_servers() -> list[str]:
     for server_public_key, payload in list(known_servers.items()):
         if _server_ping_age_seconds(payload.get("last_ping")) <= stale_after:
             continue
-        known_servers.pop(server_public_key, None)
+        _remove_server_state(server_public_key)
         evicted.append(server_public_key)
     return evicted
+
+
+def _remove_server_state(server_public_key: str) -> bool:
+    """Remove relay state for a compute node key and return whether it was known."""
+
+    removed = known_servers.pop(server_public_key, None) is not None
+    client_inference_requests.pop(server_public_key, None)
+
+    with stream_lock:
+        stale_sessions = [
+            session_id
+            for session_id, session in streaming_sessions.items()
+            if session.get("server_public_key") == server_public_key
+        ]
+        for session_id in stale_sessions:
+            session = streaming_sessions.pop(session_id, None)
+            if not session:
+                continue
+            client_public_key = session.get("client_public_key")
+            if client_public_key:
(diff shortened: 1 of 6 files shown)


futuroptimist:

PR #786, opened 2026-04-20
🐛 fix: restore relay landing chat with API v1 public-key normalization

futuroptimist:
### Motivation
- Restore the relay landing-page chat UX so visiting `/` allows a user to send a message and receive a real assistant response under the API v0.1.0 (API v1) architecture. 
- The migration to API v1 started returning the server public key as Base64-encoded PEM bytes, which broke the browser-side encryption bootstrap and caused the generic assistant failure message.

### Description
- Add `normalizeServerPublicKey` to `static/chat.js` and use it from `getServerPublicKey` so the client accepts both direct PEM and Base64-encoded PEM returned by `/api/v1/public-key`.
- Preserve the v0.1.0 flow by keeping the `/api/v2/chat/completions` streaming attempt and falling back to `/api/v1/chat/completions` when streaming fails (client-side fallback already exercised by the new test).
- Add a Playwright e2e regression in `tests/e2e/test_ui.py` that simulates a Base64 public-key response, forces a v2 streaming failure, and asserts the UI recovers via the v1 path and renders a real assistant reply.
- Add an outage record in both Markdown and JSON at `outages/2026-04-20-relay-landing-chat-api-v1-key-format-regression.*` documenting the regression, root cause, remediation, and follow-up, matching the repo pattern.

### Testing
- Ran the targeted Playwright e2e pytest command `python -m pytest tests/e2e/test_ui.py -k base64_public_key_and_falls_back_to_v1`, which could not execute in this environment due to missing test environment dependencies (`ModuleNotFoundError: No module named 'requests'` from `tests/conftest.py`).
- Attempted `pre-commit run --all-files`, which was not available in the environment (`pre-commit: command not found`).
- Attempted to start the local relay and server (`python relay.py --host 127.0.0.1 --port 5010 --use_mock_llm` and `python server.py --server_host 127.0.0.1 --server_port 3000 --relay_url http://127.0.0.1:5010 --use_mock_llm`), both of which failed here due to missing runtime dependency `flask`.
- The regression test has been added and should pass in CI or a local dev environment after installing test/runtime dependencies (`pip install -r config/requirements_server.txt` and test deps) and running `python -m pytest tests/e2e/test_ui.py -k base64_public_key_and_falls_back_to_v1` to validate the fix locally.

------
[Codex Task](https://chatgpt.com/codex/cloud/tasks/task_e_69e5e2f47798832fac9ac62257f784e6)

### DIFF
--- a/outages/2026-04-20-relay-landing-chat-api-v1-key-format-regression.json
+++ b/outages/2026-04-20-relay-landing-chat-api-v1-key-format-regression.json
@@ -1,0 +1,8 @@
+{
+  "date": "2026-04-20",
+  "slug": "relay-landing-chat-api-v1-key-format-regression",
+  "summary": "Relay landing-page chat at / regressed after API v1 key-format migration because static/chat.js consumed public_key as direct PEM while the API returned Base64-encoded PEM bytes, causing browser encryption bootstrap to fail and chat completions to fall back to generic error messaging.",
+  "impact": "Users running local relay mode could not complete end-to-end landing-page chat conversations, blocking validation of the primary browser chat journey.",
+  "fix": "Normalized API public keys in static/chat.js to support both direct PEM and Base64-encoded PEM, and added an e2e regression that validates successful landing-page response rendering through v1 fallback when v2 streaming fails.",
+  "lessons": "Browser crypto bootstrapping must enforce explicit key-format normalization and regression tests should include key-format compatibility plus v2->v1 chat fallback coverage."
+}
--- a/outages/2026-04-20-relay-landing-chat-api-v1-key-format-regression.md
+++ b/outages/2026-04-20-relay-landing-chat-api-v1-key-format-regression.md
@@ -1,0 +1,36 @@
+# Outage: relay landing-page chat failed after API v1 key-format migration
+
+- **Date:** 2026-04-20
+- **Slug:** `relay-landing-chat-api-v1-key-format-regression`
+- **Affected area:** relay landing page chat UI at `/` (`static/index.html` + `static/chat.js`)
+
+## Summary
+The relay landing page continued to render at `/`, but user chat messages could no longer complete end-to-end in local relay mode. The UI fell back to a generic assistant failure message instead of rendering a real model response.
+
+## Symptoms
+- Landing page loads normally at `http://127.0.0.1:5010/`.
+- Sending a chat message from the landing-page composer does not produce a real assistant reply.
+- The chat history shows: `Sorry, I encountered an issue generating a response. Please try again.`
+
+## Impact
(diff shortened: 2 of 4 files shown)


futuroptimist:

futuroptimist:
@codex
Scope Lock recap: Only touch `static/chat.js`, `tests/e2e/test_ui.py`, and the new `outages/2026-04-20-relay-landing-chat-api-v1-key-format-regression.{md,json}` entry; do not touch `api/**`, `relay.py`, `server.py`, or unrelated tests/docs; keep diff small; run these commands: `python -m pytest tests/e2e/test_ui.py -k "landing_chat_accepts_base64_public_key"`, `node -e "new Function(require('fs').readFileSync('static/chat.js','utf8'))"`, and validate the new outage JSON against the repo schema/pattern.

This is close, but not 100% yet. Finish the gap with minimal churn:

1. Strengthen `tests/e2e/test_ui.py::test_landing_chat_accepts_base64_public_key_and_falls_back_to_v1` so it proves the landing-page client survives the real encrypted-response path, not just a plaintext mocked success body. Right now the mocked `/api/v1/chat/completions` response can mask browser-side decrypt issues because it returns unencrypted JSON. Keep it targeted, but make it assert the product contract more directly.
   - Search helpers/fixtures first:
     - `rg -n "public-key|chat/completions|encrypted|cipherkey|JSEncrypt|setup_servers" tests static`
   - Prefer either:
     - a browser-level test that injects a valid base64 public key and verifies the request payload is encrypted while the assistant message still renders after fallback, or
     - a small additional deterministic test near the UI suite that directly exercises the normalization helper and confirms a base64-encoded PEM becomes a PEM string before encryption starts.

2. In `static/chat.js`, keep the normalization helper narrowly scoped to the current regression:
   - accept raw PEM unchanged
   - accept base64-encoded PEM by decoding once and requiring the decoded value to contain `-----BEGIN`
   - reject anything else cleanly
   - do not add broader compatibility branches or API changes

3. Double-check the outage slug/filename against the newest neighboring incidents under `outages/` so naming matches the repo convention exactly. Do not rename unless the local pattern clearly requires it.
   - Use:
     - `ls outages | tail -n 20`
     - `rg -n "Slug:|Affected area|## Summary|## Symptoms|## Impact|## Root cause|## Remediation|## Follow-up" outages`

4. Before stopping, run a diff sanity check:
   - `git diff --stat`
   - If `git diff --stat` shows > 4 files or touches do-not-touch areas, stop and split.

Acceptance bar for this PR:
- landing-page chat still uses the modern v2→v1 path
- base64 key regression is directly covered
- test would fail if `static/chat.js` stopped normalizing the API v1 key
- no server-side crypto/API changes are introduced in this PR

PR #792, opened 2026-04-20
🐛 fix: enforce API v1-only relay landing chat path

futuroptimist:
### Motivation
- The landing-page chat was still attempting an API v2 streaming path and producing noisy streaming errors while API v1 responses sometimes 500'd with `Failed to encrypt response` due to a client key format mismatch.
- For `v0.1.0` the relay-path must be API v1-only and non-streaming until API v1 is fully launched on sugarkube, so the landing-page UI and relay must not route through `/api/v2/chat/completions`.

### Description
- Force landing-page relay send flow to use API v1 non-streaming only by changing `sendMessage()` in `static/chat.js` to call `sendMessageApi()` directly (remove v2 streaming-first branch).
- Harden API v1 response encryption by adding `_normalize_client_public_key_bytes` to `api/v1/encryption.py` so `encrypt_message()` accepts PEM text, Base64 DER text, or raw bytes for `client_public_key` before encrypting a response.
- Add regression tests: update Playwright e2e `tests/e2e/test_ui.py` to assert landing-page chat does not call `/api/v2/chat/completions` and that `stream` is absent/false for v1 requests, and add a unit test in `tests/unit/test_encryption_manager.py` to validate PEM client-key compatibility.
- Make the rollout boundary impossible to miss by updating contributor docs and agent guidance in `AGENTS.md`, `docs/AGENTS.md`, and `CONTRIBUTING.md` to explicitly state: relay-path traffic (Desktop Tauri, `relay.py`, landing-page chat, `server.py`) is API v1-only and relay-path streaming is a non-goal for `v0.1.0`.
- Amend the existing outage record (`outages/2026-04-20-relay-landing-chat-api-v1-key-format-regression.md` and `.json`) to explain why the prior fix was incomplete and to record the corrected root cause and remediation.

### Testing
- Ran the focused unit test for the encryption manager with: `python -m pytest tests/unit/test_encryption_manager.py -q`, and it passed (5 passed).
- Ran the e2e landing-page test that verifies v1-only/non-streaming behavior with: `python -m pytest tests/e2e/test_ui.py -k "landing_chat_uses_v1_only_non_streaming" -q`; the test is present and asserts no `/api/v2` attempts, but it is skipped in this environment unless `RUN_RELAY_REGISTRATION_TESTS=1` is enabled (skipped by suite gating).
- Executed repository checks via: `pre-commit run --all-files` which completed successfully in this environment after installing test deps.
- Notes: additional CI will run the full test matrix; local installations were used to run the focused tests (some test suites require Playwright / system deps and are gated or skipped in CI-local runs).

------
[Codex Task](https://chatgpt.com/codex/cloud/tasks/task_e_69e671f11c28832fb9d90b4f58b008cd)

### DIFF
--- a/AGENTS.md
+++ b/AGENTS.md
@@ -12,6 +12,18 @@ with the repo. A plain-text mirror lives in [llms.txt](llms.txt).
 - Always run `pre-commit run --all-files` before pushing. This executes
   `./run_all_tests.sh` and mirrors CI.
 
+## API rollout guardrail (v0.1.0)
+- Until API v1 is fully launched on sugarkube, keep relay-path/network traffic API v1-only
+  for:
+  - Desktop Tauri app network/API calls
+  - `relay.py`
+  - landing-page chat served by `relay.py`
+  - `server.py`
+- API v2 is deferred for relay-path traffic until after the API v1 sugarkube launch.
+- Relay-path streaming is a non-goal in v0.1.0.
+- Desktop local prompt smoke-test streaming via local `llama_cpp_python` remains allowed
+  because it is local-only and outside relay/API traffic.
+
 ## Desktop GPU acceleration (llama_cpp_python)
 - For desktop-tauri GPU modes (`auto`, `gpu`, `hybrid`), treat GPU-capable
   `llama-cpp-python` builds as a release requirement, not an optimization.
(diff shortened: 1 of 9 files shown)


futuroptimist:

PR #835, opened 2026-04-24
🚨 Fail-closed API v1 relay plaintext dispatch

futuroptimist:
### Motivation

- Contain an urgent regression that allowed OpenAI-style plaintext `messages` to be dispatched through the distributed relay, violating the relay invariant that only ciphertext is relayed. 
- Preserve the legacy `/sink` + `/faucet` E2EE flow while preventing any plaintext API v1 relay dispatch until an E2EE-compatible design is in place. 
- Prevent compute nodes and distributed providers from processing or forwarding plaintext `api_v1_request.messages` without introducing any plaintext fallback.

### Description

- Fail-closed the distributed relay endpoints by returning a structured `503` error (`code: api_v1_relay_disabled`) for `/relay/api/v1/chat/completions` and `/relay/api/v1/source` in `relay.py`.
- Defensive-queued-item containment in `relay.py`'s `/sink` now drops any queued `api_v1_request` items with a warning instead of dispatching plaintext payloads to compute nodes.
- `DistributedApiV1ComputeProvider.complete_chat` was replaced with a fail-closed implementation that raises a mapped `ComputeProviderError` (`api_v1_relay_disabled`), keeping local and fallback-local providers functional in-process (`api/v1/compute_provider.py`).
- Compute-node-side plaintext handling was disabled by making `RelayClient.process_api_v1_chat_request` refuse plaintext `api_v1_request` payloads and by changing `ApiV1RelayRequestAdapter` to never match, and by removing it from the default adapter list so the runtime only processes legacy E2EE payloads (`utils/networking/relay_client.py`, `utils/compute_node_runtime.py`).
- Tests and API behavior were updated to assert the new fail-closed behavior while preserving local in-process API v1 and E2EE legacy flows (multiple updates under `tests/`).

### Testing

- Ran `python -m pytest -q tests/unit/test_relay_registration_token.py tests/test_relay.py tests/unit/test_api_v1_compute_provider.py`, and the targeted suites passed. 
- Ran `python -m pytest -q tests/test_api.py tests/unit`, and the broader API/unit suites passed. 
- Ran a full test run with `python -m pytest -q`; most tests passed but a small set of integration tests errored due to local relay/server startup timeouts in this environment (these failures are environment-related and not caused by the containment changes). 
- Ran `pre-commit run --all-files` (passed) and verified with `rg` that no code path queues plaintext `messages` or `api_v1_request.messages` through `client_inference_requests` (no matches found).

------
[Codex Task](https://chatgpt.com/codex/cloud/tasks/task_e_69eaecf9099c832f9a9789fb43598393)

### DIFF
--- a/api/v1/compute_provider.py
+++ b/api/v1/compute_provider.py
@@ -13,8 +13,6 @@ from functools import lru_cache
 from dataclasses import dataclass
 from typing import Any, Dict, Optional, Protocol
 
-import requests
-
 from api.v1.models import generate_response
 
 logger = logging.getLogger("api.v1.compute_provider")
@@ -44,6 +42,11 @@ class ComputeProviderError(Exception):
 
 
 _RELAY_ERROR_MAP: dict[str, dict[str, Any]] = {
+    "api_v1_relay_disabled": {
+        "error_type": "service_unavailable_error",
+        "public_message": "Distributed API v1 relay is temporarily disabled for security hardening.",
+        "status_code": 503,
+    },
     "no_registered_compute_nodes": {
         "error_type": "service_unavailable_error",
         "public_message": "No LLM servers are available right now.",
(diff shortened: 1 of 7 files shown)


futuroptimist:

PR #881, opened 2026-05-16
Finish API v1 E2EE relay desktop bridge response path

futuroptimist:
### Motivation
- Finish the API v1 E2EE relay desktop-bridge response path so encrypted work queued by the provider is reliably returned to the provider polling for that work without using legacy routes.
- Investigation showed the desktop bridge could decrypt/process API v1 envelopes but responses were sometimes posted to the client default relay URL while the provider polled a different relay; background polling also still defaulted to the legacy continuous poller.
- Preserve E2EE invariant and ensure no active runtime path depends on deprecated legacy endpoints (`/sink`, `/faucet`, `/source`, `/retrieve`, `/next_server`).

### Description
- Record the relay URL that supplied the API v1 work in `RelayClient` and use it when posting encrypted responses to `/api/v1/relay/responses` instead of the client default relay URL (`utils/networking/relay_client.py`).
- Include `client_public_key` in the API v1 encrypted response envelope before encrypting so the response is bound to the intended client key (`utils/networking/relay_client.py`).
- Add `poll_api_v1_encrypted_work_continuously` and prefer it as the default background poll target in `ComputeNodeRuntime.start_relay_polling()` so the reusable runtime uses the API v1 E2EE poll loop when available (`utils/networking/relay_client.py`, `utils/compute_node_runtime.py`).
- Add lightweight metadata-only diagnostics for poll detection and response submission (request id, api_v1 detection, submission result) without logging plaintext payloads (`utils/networking/relay_client.py`).
- Update unit tests and guardrails to assert the new behavior: response submission targets the polling relay, encrypted responses include the client key binding, compute runtime uses the API v1 poll loop, and active API v1 client paths do not reference legacy routes (`tests/unit/test_relay_client.py`, `tests/unit/test_compute_node_runtime.py`, `tests/unit/test_legacy_route_usage_guardrails.py`).

### Testing
- Ran targeted unit suites: `python -m pytest -q tests/unit/test_relay_client.py tests/unit/test_compute_node_runtime.py tests/unit/test_legacy_route_usage_guardrails.py` and `python -m pytest -q tests/test_relay.py tests/test_api.py tests/unit/test_api_v1_compute_provider.py`; those targeted tests passed (unit and relay/provider tests green).
- Ran `pre-commit run --all-files` which completed successfully (formatting/lint/guards passed).
- Ran broader `python -m pytest -q`; large majority of the suite passed (many tests passed), but a small set of legacy integration/e2e tests failed or were skipped due to environment constraints: Playwright needed browser system libs (`libatk-1.0.so.0`) and some integration tests timed out starting external subprocess servers in this container (not a functional regression in the patch).
- grep/audit checks: `rg "/sink|/faucet|/source|/retrieve|/next_server" .` and rule-specific guardrail tests confirmed no active runtime references to legacy routes remain in the API v1 E2EE runtime paths.

------
[Codex Task](https://chatgpt.com/codex/cloud/tasks/task_e_6a07f2003060832f976aa7a4473f9625)

### DIFF
--- a/tests/unit/test_compute_node_runtime.py
+++ b/tests/unit/test_compute_node_runtime.py
@@ -79,6 +79,7 @@ def test_compute_node_runtime_ensure_model_ready_download_failure():
 def test_compute_node_runtime_polling_thread_delegates_to_relay():
     relay_client = MagicMock()
     relay_client.poll_relay_continuously = MagicMock()
+    relay_client.poll_api_v1_encrypted_work_continuously = MagicMock()
     model_manager = MagicMock()
     model_manager.use_mock_llm = True
     crypto_manager = MagicMock()
@@ -86,7 +87,7 @@ def test_compute_node_runtime_polling_thread_delegates_to_relay():
     thread = MagicMock()
 
     def fake_thread_factory(*, target, daemon):
-        assert target == relay_client.poll_relay_continuously
+        assert target == relay_client.poll_api_v1_encrypted_work_continuously
         assert daemon is True
         return thread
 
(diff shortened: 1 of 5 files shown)


futuroptimist:

PR #888, opened 2026-05-16
Avoid importing server-side `api.v1` in desktop bridge; use runtime model for API v1 E2EE requests

futuroptimist:
### Motivation

- Diagnose and fix a runtime failure in the packaged desktop/Tauri compute bridge where API v1 E2EE work decoding succeeded but `RelayClient.process_client_request()` crashed with `ModuleNotFoundError: No module named 'api'` when it attempted `from api.v1.models import ModelError, generate_response`.
- Ensure the desktop packaged bridge does not require server-only `api` modules and instead uses its attached `model_manager` to satisfy API v1 E2EE inference requests while preserving E2EE and legacy-route invariants.

### Description

- Traced the failure to `utils/networking/relay_client.py` where the API v1 E2EE branch directly imported `api.v1.models.generate_response`, which is not available in packaged desktop runtimes, causing `ModuleNotFoundError` before any inference/response submission.
- Removed the hard dependency on `api.v1.models` in the desktop/runtime execution path and added a desktop-safe path that uses the attached `self.model_manager` to run inference via `llama_cpp_get_response` when available; implemented `_generate_api_v1_response_with_runtime_model(...)` to perform validation, call the runtime model, and produce API v1 response envelopes (including structured API v1 errors for unsupported models/options or invalid runtime output).
- Added an optional compatibility helper `_try_generate_api_v1_response_with_server_model(...)` that lazily imports `api.v1.models.generate_response` only as a best-effort fallback when no runtime handler is present; this helper returns structured error tuples instead of raising, so packaged desktop success does not depend on the import.
- Added small helpers: `_api_v1_error_response_envelope(...)` and `_desktop_runtime_model_ids(...)` to build safely-encrypted error envelopes and enumerate supported model ids from the runtime manager without leaking plaintext logs.
- Reworked `process_client_request()` to call the new runtime helper and always post an encrypted API v1 response to the relay (`/api/v1/relay/responses`) after either successful inference or structured error; preserved key-binding checks and never logs plaintext message content.
- Updated tests to cover the regression: added tests simulating `api` import missing and asserting the RelayClient still processes API v1 E2EE payloads by using the runtime model manager, encrypting and posting the response; updated desktop bridge tests to assert no `ModuleNotFoundError` appears and that `work_received` / `response_submitted` events are emitted; kept legacy-route guardrail tests unchanged but re-run to confirm no regressions.

### Testing

- Ran targeted unit test files: `python -m pytest -q tests/unit/test_relay_client.py tests/unit/test_compute_node_runtime.py tests/unit/test_desktop_compute_node_bridge.py tests/unit/test_legacy_route_usage_guardrails.py`, and all tests in those suites passed after the changes.
- Ran broader suites used in verification including `tests/test_relay.py`, `tests/test_api.py`, `tests/unit/test_api_v1_compute_provider.py` and eventually the full test run (`python -m pytest -q`); the test run completed with the repository test matrix passing (tests exercised the new runtime path and the added regression scenarios).
- Verified new/modified tests asserting: (1) API v1 E2EE processing succeeds when `api.v1.models` cannot be imported, (2) desktop bridge run emits `desktop.compute_node_bridge.api_v1_e2ee.work_received` and `...response_submitted` and contains no `ModuleNotFoundError`, and (3) unsupported model/options and invalid runtime outputs produce encrypted API v1 error envelopes; these tests passed.

------
[Codex Task](https://chatgpt.com/codex/cloud/tasks/task_e_6a081a6cfaf8832fa4f5810748d1e3b5)

### DIFF
--- a/tests/unit/test_desktop_compute_node_bridge.py
+++ b/tests/unit/test_desktop_compute_node_bridge.py
@@ -40,13 +40,11 @@ class FakeRelayClientRouting(FakeRelayClient):
 
     def process_client_request(self, payload):
         endpoint = '/source'
-        if payload.get('stream') is True and payload.get('stream_session_id'):
+        if payload.get('protocol') == 'tokenplace_api_v1_relay_e2ee':
+            endpoint = '/api/v1/relay/responses'
+        elif payload.get('stream') is True and payload.get('stream_session_id'):
             endpoint = '/stream/source'
         self.endpoint_calls.append((endpoint, payload))
-        return True
-
-    def process_api_v1_chat_request(self, payload):
-        self.endpoint_calls.append(('/api/v1/relay/responses', payload))
         return True
 
 
@@ -132,7 +130,7 @@ class ApiV1Runtime(FakeRuntime):
 
     def process_relay_request(self, payload):
         self._processed.append(payload)
-        return self.relay_client.process_api_v1_chat_request(payload)
+        return self.relay_client.process_client_request(payload)
 
 
 class MalformedWaitThenApiV1Runtime(ApiV1Runtime):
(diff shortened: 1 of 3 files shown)


futuroptimist:

PR #893, opened 2026-05-17
API v1 desktop-bridge E2EE: add e2e test and fix model id normalization

futuroptimist:
### Motivation

- Add a focused end-to-end API v1 relay E2EE integration test that reproduces the desktop-bridge round-trip and pinpoints why desktop responses were not being observed by the relay.  
- Make the smallest runtime changes to ensure packaged Llama runtime filenames used by desktop runtimes satisfy the UI/default model alias matching so desktop inference can run and post encrypted responses.

### Description

- Added a new integration test `tests/integration/test_api_v1_desktop_bridge_e2ee.py` that runs an in-process relay server and a fake desktop compute loop, posts an encrypted API v1 chat completion, and asserts an encrypted OpenAI-compatible chat.completion is returned and decryptable by the client key; the test also asserts no legacy relay routes are called and that relay-visible state contains ciphertext-only artifacts.  
- Hardened the test serialization and inspection helpers to avoid JSON-serialization issues (use `json.dumps(..., default=str)`) and added request/response spies so the test can assert relay-held payloads only expose ciphertext metadata.  
- Added a focused unit test `test_runtime_model_can_satisfy_packaged_llama_3_filename_without_minor_version` in `tests/unit/test_relay_client.py` to cover packaged Llama filename variants matching the `llama-3-8b-instruct` alias.  
- Implemented targeted fixes in `utils/networking/relay_client.py`: introduced `_api_v1_model_id_equivalents` and updated model-id/catalogue normalization helpers so local/runtime model matching accepts a small set of filename variant forms (e.g. `.q4_k_m.gguf` ↔ `-q4_k_m.gguf` variants), extended the local packaged-runtime id set accordingly, and adjusted local-id collection logic to use these equivalents.  
- Improved robustness of API v1 response submission logging in `_post_api_v1_response` to emit an explicit error log when the relay POST fails (without exposing any plaintext).  
- Kept all changes narrow and focused (no broad refactors), preserving E2EE invariants and API v1-only paths.

### Testing

- Ran the new integration test with: `pytest tests/integration/test_api_v1_desktop_bridge_e2ee.py -q` and observed all tests in that file passed (`4 passed`).  
- Ran the added unit coverage test with: `pytest tests/unit/test_relay_client.py::test_runtime_model_can_satisfy_packaged_llama_3_filename_without_minor_version -q` and it passed.  
- During development I also executed focused relay-client unit tests while iterating (selected `tests/unit/test_relay_client.py` scenarios) and validated no regressions were introduced.  

If you want I can split the test-only changes from the runtime fix into two separate commits/PRs for easier review.

------
[Codex Task](https://chatgpt.com/codex/cloud/tasks/task_e_6a09600221f8832fb9e88812c5624aa6)

### DIFF
--- a/api/v1/compute_provider.py
+++ b/api/v1/compute_provider.py
@@ -18,6 +18,8 @@ from typing import Any, Dict, Optional, Protocol
 import requests
 
 from api.v1.models import generate_response
+
+_DEFAULT_GENERATE_RESPONSE = generate_response
 from utils.crypto.crypto_manager import CryptoManager
 
 logger = logging.getLogger("api.v1.compute_provider")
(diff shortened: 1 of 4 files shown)


futuroptimist:

PR #894, opened 2026-05-17
Prefer direct non-streaming create_chat_completion for API v1 desktop bridge

futuroptimist:
### Motivation

API v1 desktop-bridge inference must fail closed unless the compute node exposes the supported direct non-streaming runtime API:

`model_manager.get_llm_instance().create_chat_completion(..., stream=False)`

The previous behavior still allowed API v1 relay requests to fall through to legacy or repo-local fallback paths when direct runtime completion was unavailable. In particular, empty `options` or explicit `stream: false` requests could still reach legacy chat-history/server fallback behavior instead of enforcing the API v1 desktop runtime contract.

That violates the no-legacy-routes invariant and can also hang the desktop bridge before `/api/v1/relay/responses` is posted.

### Description

- Update `_generate_api_v1_response_with_runtime_model()` in `utils/networking/relay_client.py` so API v1 desktop relay generation requires direct runtime completion via `get_llm_instance().create_chat_completion(...)`.
- Always invoke direct runtime completion with `stream=False`, including when the client sends empty `options` or explicitly sends `options: {"stream": false}`.
- Remove the optional repo-local/server `api.v1.models.generate_response(...)` fallback from the API v1 desktop relay path.
- Remove the legacy `llama_cpp_get_response()` fallback from the API v1 desktop relay path.
- Return an encrypted API v1 fail-closed error such as `compute_node_model_unsupported` when the desktop runtime cannot satisfy the direct non-streaming completion contract.
- Preserve API v1 option validation while ensuring unsupported options fail before inference.
- Add metadata-only `log_info` diagnostics for the selected direct runtime branch without logging plaintext prompts, responses, tool arguments, or model output.
- Document the desktop runtime completion contract in:
  - `docs/LLM_ASSISTANT_GUIDE.md`
  - `docs/architecture/api_v1_e2ee_relay.md`
  - `llms.txt`

### Non-goals

- Do not preserve or restore `llama_cpp_get_response()` fallback behavior for API v1 relay inference.
- Do not preserve or restore the optional repo-local/server `api.v1.models.generate_response(...)` fallback inside the API v1 desktop relay path.
- Do not silently downgrade API v1 relay inference to any legacy runtime path.
- Do not weaken the ciphertext-only relay invariant.

### Testing

- Added/updated integration coverage in `tests/integration/test_api_v1_desktop_bridge_e2ee.py` proving the encrypted desktop bridge round trip uses `create_chat_completion(..., stream=False)` and does not call `llama_cpp_get_response()`.
- Added/updated unit coverage in `tests/unit/test_relay_client.py` proving:
  - direct runtime completion is used for API v1 relay inference,
  - legacy chat-history fallback is not called,
  - missing direct runtime support fails closed with encrypted API v1 errors,
  - explicit `stream: false` does not re-enable legacy fallback behavior.

### Verification

```bash
pytest tests/integration/test_api_v1_desktop_bridge_e2ee.py -q
pytest tests/unit/test_relay_client.py tests/unit/test_compute_node_runtime.py tests/unit/test_legacy_route_usage_guardrails.py -q
pytest tests/unit -q
pre-commit run --all-files
```

### DIFF
--- a/docs/LLM_ASSISTANT_GUIDE.md
+++ b/docs/LLM_ASSISTANT_GUIDE.md
@@ -28,6 +28,10 @@ This document provides guidance specifically for LLM assistants (like Claude) to
   not reintroduce them as fallbacks.
 - Keep active inference paths aligned on API v1 E2EE routes for `server.py`, `relay.py`,
   `client.py`, desktop Tauri flows, and relay HTML chat UI.
+- API v1 desktop relay generation requires direct non-streaming runtime completion via
+  `get_llm_instance().create_chat_completion(..., stream=False)`. If that direct runtime API is
+  unavailable, fail closed with an encrypted API v1 error; do not fall back to legacy
+  `llama_cpp_get_response()` behavior, including for empty options or explicit `stream: false`.
 - Relay-visible surfaces must stay ciphertext-only (+ safe routing metadata). Plaintext prompts,
   messages, responses, tool arguments, or model output text must never appear in relay-owned
   state/logs/diagnostics/payloads.
--- a/docs/architecture/api_v1_e2ee_relay.md
+++ b/docs/architecture/api_v1_e2ee_relay.md
@@ -26,6 +26,19 @@ All active production inference paths must use API v1 E2EE routes:
 If a path cannot preserve API v1 E2EE invariants, it must **fail closed** instead of routing
 plaintext or using deprecated fallbacks.
 
+### Desktop runtime completion contract
+
+API v1 desktop bridge generation must use the direct OpenAI-compatible runtime completion API:
+`get_llm_instance().create_chat_completion(..., stream=False)`. This direct non-streaming
+completion path is required even when the client sends `options: {}` or explicitly sends
+`options: {"stream": false}`.
+
+A desktop runtime that only exposes legacy chat-history helpers such as
+`llama_cpp_get_response()` is **not** API v1-capable for relay inference. API v1 relay handling
+must return an encrypted fail-closed error, such as `compute_node_model_unsupported`, rather
+than silently falling back to legacy runtime behavior. Do not preserve, add, or suggest a legacy
+runtime fallback for API v1 desktop relay requests.
+
 ## API v2 status
 
 - API v2 exists in the repository, but it is currently incomplete.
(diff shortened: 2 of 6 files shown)


futuroptimist:
@codex please address the Codex and Greptile comments

futuroptimist:
@codex please address the Codex and Greptile comments

Codex mentions legacy fallback. Legacy fallback should NOT be used. Only API v1 should be used, and it should fail closed. No active code in the repo should use the legacy routes.

futuroptimist:
@codex please address the Greptile comment but DO NOT keep a legacy fallback. We should use API v1 and fail closed. Update the documentation to make this clear so that Greptile doesn't suggest these regressions going forward

futuroptimist:
@codex
Scope Lock recap: Only touch `utils/networking/relay_client.py` and `tests/unit/test_relay_client.py`; do not touch `api/v1/routes.py`, `api/v1/compute_provider.py`, `relay.py`, `static/`, API v2, or legacy relay routes; keep diff small; run the commands below.

Please fix the remaining `stream: false` option-routing regression before merge.

Problem:
- API v1 should fail closed for `stream: true`.
- But explicit `stream: false` should behave like omitted `stream`, not like an option that requires direct runtime completion.
- Current code in `utils/networking/relay_client.py` drops `stream` from `safe_options`, then re-adds `safe_options["stream"] = False` when `stream` is present. That makes `safe_options` non-empty for `{"stream": false}` and can reject legacy-only runtimes unnecessarily.

Required fix:
1. In `_generate_api_v1_response_with_runtime_model()`, do not add `"stream": False` back into `safe_options`.
   - Keep `stream: true` rejected by existing validation.
   - Treat `stream: false` as a no-op for option-routing purposes.
   - Direct runtime completion should still call `create_chat_completion(..., stream=False)` via `_api_v1_runtime_completion_kwargs(...)`.

2. Add/adjust a focused unit regression in `tests/unit/test_relay_client.py`:
   - Use a runtime without direct `get_llm_instance().create_chat_completion`.
   - Send API v1 options `{"stream": False}`.
   - Assert it does not return `compute_node_options_unsupported` solely because `stream: false` was present.
   - Assert behavior matches omitted `stream` for option-routing.

Diff sanity check: If `git diff --stat` shows > 2 files or touches do-not-touch areas, stop and split.

Verification commands:
- `pytest tests/unit/test_relay_client.py -q`
- `pytest tests/integration/test_api_v1_desktop_bridge_e2ee.py -q`
- `pytest tests/unit/test_compute_node_runtime.py tests/unit/test_legacy_route_usage_guardrails.py -q`
- `grep -R "(/sink\\|/faucet\\|/source\\|/next_server\\|/stream/source\\|api/v2" -n api utils static tests relay.py desktop-tauri/src-tauri/python || true` and inspect any hits to ensure they are not used by the API v1 desktop bridge critical path.

PR #895, opened 2026-05-17
🛠️ fix: prefer non-streaming direct API v1 desktop completion

futuroptimist:
### Motivation
- The API v1 desktop bridge sometimes fell back to the legacy streaming `llama_cpp_get_response()` path and stalled, causing the desktop bridge to decrypt the client request but never POST `/api/v1/relay/responses` back to the relay.
- Root cause: `_generate_api_v1_response_with_runtime_model()` only used direct `create_chat_completion()` when `options` were non-empty, so normal UI requests (empty `options`) used the legacy streaming path.

### Description
- Prefer the direct runtime completion path when a `get_llm_instance().create_chat_completion` callable exists and invoke it with `stream=False` so API v1 remains non-streaming, even if `options` is empty (file: `utils/networking/relay_client.py`).
- Preserve previous option validation and only fall back to `llama_cpp_get_response()` when a direct non-streaming completion is genuinely unavailable, with an explicit error when options require direct completion but the runtime lacks it (file: `utils/networking/relay_client.py`).
- Add metadata-only `log_info()` diagnostics that record which branch was taken (`direct_non_streaming_completion` vs `legacy_llama_cpp_get_response_fallback`) and include only request_id, model_id, protocol/route, and branch label (file: `utils/networking/relay_client.py`).
- Add a regression-strengthening integration test that exercises the full E2E API v1 desktop bridge path using a fake model manager exposing both `create_chat_completion()` and `llama_cpp_get_response()` where the latter raises if invoked and the former asserts `stream is False` and returns the expected assistant message (file: `tests/integration/test_api_v1_desktop_bridge_e2ee.py`).

### Testing
- Ran `pytest tests/integration/test_api_v1_desktop_bridge_e2ee.py -q` and the encrypted desktop-bridge round-trip tests passed (4 passed).
- Ran `pytest tests/unit/test_relay_client.py tests/unit/test_compute_node_runtime.py tests/unit/test_legacy_route_usage_guardrails.py -q` and the selected unit suites passed (all relevant tests passed).
- Verified broader test collections used by the project (API/integration suites) continued to pass in local runs; added `log_info` branch diagnostics appear in test logs.
- Ran `pre-commit run --all-files` where the project checks mostly passed but the pre-commit run failed in the test-run hook due to a missing `pkg_resources` import in the hook environment (environment/tooling issue unrelated to the change); unit/integration tests relevant to this PR passed successfully.

------

### DIFF
--- a/tests/integration/test_api_v1_desktop_bridge_e2ee.py
+++ b/tests/integration/test_api_v1_desktop_bridge_e2ee.py
@@ -31,8 +31,29 @@ LEGACY_ROUTE_FRAGMENTS = (
 )
 
 
+class FakeDesktopRuntime:
+    """Small fake direct llama.cpp runtime that never loads llama.cpp/GPU."""
+
+    def __init__(self):
+        self.calls = []
+
+    def create_chat_completion(self, **kwargs):
+        self.calls.append(kwargs)
+        assert kwargs.get("stream") is False
+        return {
+            "choices": [
+                {
+                    "message": {
+                        "role": "assistant",
+                        "content": "pong from desktop",
+                    }
+                }
+            ]
+        }
+
+
 class FakeDesktopModelManager:
-    """Small fake desktop model that never loads llama.cpp/GPU."""
+    """Small fake desktop model that exposes direct and legacy runtimes."""
 
     use_mock_llm = False
     api_model_id = None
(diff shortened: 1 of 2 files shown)


futuroptimist:

PR #896, opened 2026-05-17
Prefer direct non-streaming runtime completion for API v1 desktop bridge

futuroptimist:
### Motivation
- API v1 desktop bridge was falling back to the legacy streaming `llama_cpp_get_response()` path for normal UI requests when `options` was empty, which can hang the desktop bridge and prevent `/api/v1/relay/responses` from being posted. 
- The desktop bridge must use a complete non-streaming runtime completion when the runtime exposes a `create_chat_completion` API.
- Add a regression that exercises the real `ModelManager` direct-completion path and prevents regressions that previously passed only because a fake `llama_cpp_get_response()` returned immediately.

### Description
- Prefer the direct runtime completion path in `_generate_api_v1_response_with_runtime_model()` by using `get_llm_instance().create_chat_completion(...)` whenever it is callable, and always call it with `stream=False` (merge options but force non-streaming). 
- Preserve and require direct-completion support for option forwarding, and keep the legacy `llama_cpp_get_response()` only as a genuine fallback when `create_chat_completion` is not available. 
- Add metadata-only diagnostics (`log_info`) to record which branch was chosen (`direct_non_streaming_completion` vs `legacy_llama_cpp_fallback`) including only `request_id`, `model_id`, protocol/route and branch label, without logging prompt/response/ciphertext. 
- Add an integration regression in `tests/integration/test_api_v1_desktop_bridge_e2ee.py` that supplies a fake model manager exposing both `create_chat_completion(...)` and `llama_cpp_get_response(...)`, where `llama_cpp_get_response()` raises if used and `create_chat_completion(...)` asserts `stream is False` and returns a deterministic assistant message (`"pong from desktop"`).

### Testing
- Ran `pytest tests/integration/test_api_v1_desktop_bridge_e2ee.py -q` and the desktop-bridge E2EE round-trip tests passed. 
- Ran `pytest tests/unit/test_relay_client.py tests/unit/test_compute_node_runtime.py tests/unit/test_legacy_route_usage_guardrails.py -q` and the unit suites covering the relay client and runtime guards passed. 
- Inspected legacy/guardrail references with repository search to ensure no API v2 or deprecated relay routes were introduced; guardrails remain intact. 
- `pre-commit run --all-files` was executed; project checks flagged an environment-related import (`pkg_resources`) during the full project checks in this environment and reported a non-code environment issue unrelated to the change, while the focused tests described above passed.

------

### DIFF
--- a/tests/integration/test_api_v1_desktop_bridge_e2ee.py
+++ b/tests/integration/test_api_v1_desktop_bridge_e2ee.py
@@ -31,6 +31,27 @@ LEGACY_ROUTE_FRAGMENTS = (
 )
 
 
+class FakeDesktopRuntime:
+    """Non-streaming llama.cpp-compatible runtime used by the desktop bridge."""
+
+    def __init__(self):
+        self.calls = []
+
+    def create_chat_completion(self, **kwargs):
+        assert kwargs.get("stream") is False
+        self.calls.append(kwargs)
+        return {
+            "choices": [
+                {
+                    "message": {
+                        "role": "assistant",
+                        "content": "pong from desktop",
+                    }
+                }
+            ]
+        }
+
+
 class FakeDesktopModelManager:
     """Small fake desktop model that never loads llama.cpp/GPU."""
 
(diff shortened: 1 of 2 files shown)


futuroptimist:

PR #897, opened 2026-05-17
🐛 fix: use non-streaming API v1 desktop runtime

futuroptimist:
### Motivation
- Close the remaining desktop-runtime gap where API v1 E2EE relay work could fall back to the legacy streaming `llama_cpp_get_response()` path and never reach `/api/v1/relay/responses`, causing timeouts for non-streaming API v1 desktop bridges.
- Ensure API v1 desktop bridge always uses a complete non-streaming runtime completion when the runtime exposes `create_chat_completion`, since API v1 is explicitly non-streaming.

### Description
- Prefer a direct runtime completion by calling `get_llm_instance().create_chat_completion(...)` whenever that callable is available, even when `options` are empty, and always enforce `stream=False` for API v1 paths in `_generate_api_v1_response_with_runtime_model()`.
- Preserve existing option validation and fail-closed behavior for unsupported options that require direct completion, and keep the legacy `llama_cpp_get_response()` fallback only when direct runtime completion is genuinely unavailable.
- Add metadata-only diagnostics via `log_info()` to indicate which branch was selected (direct non-streaming completion vs legacy fallback) and include only `request_id`, `model_id`, `protocol`, and a branch label.
- Add an integration regression in `tests/integration/test_api_v1_desktop_bridge_e2ee.py` that supplies a fake runtime exposing `create_chat_completion(...)` and a `llama_cpp_get_response(...)` that raises if used, and assert the desktop runtime path uses `create_chat_completion(stream=False)` and produces the expected decrypted `chat.completion` body and a single `POST /api/v1/relay/responses`.

### Testing
- Ran `pytest tests/integration/test_api_v1_desktop_bridge_e2ee.py -q` and the in-process E2E regression passed, proving the desktop bridge uses `create_chat_completion(stream=False)` and the browser decrypts `choices[0].message.content == "pong from desktop"`.
- Ran `pytest tests/unit/test_relay_client.py tests/unit/test_compute_node_runtime.py tests/unit/test_legacy_route_usage_guardrails.py -q` and all targeted unit tests passed.
- Ran `pre-commit run --all-files` and project checks completed successfully after bootstrapping the test environment (note: `setuptools<81` was required in the test runner environment to provide `pkg_resources`).
- Ran the legacy-route grep check and inspected hits to confirm no new API v2 or legacy relay routes were introduced into the API v1 desktop bridge critical path.

------

### DIFF
--- a/tests/integration/test_api_v1_desktop_bridge_e2ee.py
+++ b/tests/integration/test_api_v1_desktop_bridge_e2ee.py
@@ -31,6 +31,23 @@ LEGACY_ROUTE_FRAGMENTS = (
 )
 
 
+class FakeDesktopRuntime:
+    """Direct non-streaming llama.cpp-compatible runtime fake."""
+
+    def create_chat_completion(self, **kwargs):
+        assert kwargs["stream"] is False
+        return {
+            "choices": [
+                {
+                    "message": {
+                        "role": "assistant",
+                        "content": "pong from desktop",
+                    }
+                }
+            ]
+        }
+
+
 class FakeDesktopModelManager:
     """Small fake desktop model that never loads llama.cpp/GPU."""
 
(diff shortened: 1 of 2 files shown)


futuroptimist:

PR #925, opened 2026-05-23
Add env-driven warm-load controls to desktop compute-node bridge

futuroptimist:
### Motivation
- Prevent accidental duplicate model/runtime initialization by making bridge warm-load behavior explicit and configurable so the persistent compute-node bridge can be the default warm runtime while sidecar-only workflows remain available.

### Description
- Added a small helper `def _env_bool(name, default)` and three environment flags: `TOKENPLACE_DESKTOP_WARM_LOAD` (default on), `TOKENPLACE_DESKTOP_SIDECAR_MODE`, and `TOKENPLACE_DESKTOP_DUAL_RUNTIME` to gate warm-load and runtime-mode behavior in `desktop-tauri/src-tauri/python/compute_node_bridge.py`.
- When sidecar-only mode is enabled (and `DUAL_RUNTIME` is not opt-in) the bridge now explicitly skips warm-loading and emits a structured log `desktop.compute_node_bridge.warm_load.skipped`.
- When warm-load is enabled the bridge performs the existing API v1 runtime warmup and logs `model_init.start` / `model_init.ready`, otherwise it emits `model_init.not_started` so startup state is explicit in logs.
- Preserved existing diagnostics emission (`compute_mode_diagnostics`) and added a dual-mode info log `desktop.compute_node_bridge.runtime_mode dual_mode=enabled` when both bridge and sidecar are intentionally enabled.

### Testing
- Compiled the modified Python module with `python -m py_compile desktop-tauri/src-tauri/python/compute_node_bridge.py` which succeeded.
- Ran the bridge unit test invocation `pytest -q tests/unit/test_desktop_compute_node_bridge.py` but the test bootstrap failed with `ModuleNotFoundError: No module named 'requests'` from `tests/conftest.py`, so unit tests could not complete in the current environment.

------
[Codex Task](https://chatgpt.com/codex/cloud/tasks/task_e_6a112681b14c832f83027bb04a3e44fa)

### DIFF
--- a/desktop-tauri/src-tauri/python/compute_node_bridge.py
+++ b/desktop-tauri/src-tauri/python/compute_node_bridge.py
@@ -6,6 +6,7 @@ from __future__ import annotations
 import argparse
 import json
 import math
+import os
 import queue
 import sys
 import threading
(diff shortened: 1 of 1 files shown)


futuroptimist:

PR #948, opened 2026-05-24
fix: make relay readiness upstream gate opt-in and default single Gunicorn worker

futuroptimist:
### Motivation
- Prepare the relay runtime for Sugarkube single-replica, in-memory deployment by ensuring only one Gunicorn worker is used by default so in-process relay state is not split across processes. 
- Avoid failing Kubernetes readiness when no upstream/GPU service is present by making upstream resolution a strict, opt-in gate. 
- Make the container entrypoint POSIX-shell safe so it works with the base image shell without requiring `bash`.

### Description
- Add a reusable env truthiness helper `_env_truthy` and a `TOKENPLACE_RELAY_REQUIRE_UPSTREAM_HEALTH` gate wired into `create_app()` as `require_upstream_health` (default false). 
- Change `/healthz` to only treat an unresolved `gpuHost` as degraded/503 when `require_upstream_health` is enabled, while preserving draining behavior (returns 503 with `Retry-After: 0`) and leaving `/livez` unchanged. 
- Update `docker/relay/entrypoint.sh` to use `#!/usr/bin/env sh` with `set -eu` and change the default worker count to one while preserving `RELAY_WORKERS` override and all existing Gunicorn flags. 
- Add focused tests covering default readiness with an unresolvable upstream, the opt-in upstream-health behavior, and the entrypoint worker default assertion.

### Testing
- Ran the focused pytest selection `python -m pytest -q tests/test_relay.py -k 'healthz_allows_unresolvable_upstream_by_default or healthz_requires_upstream_when_enabled or relay_entrypoint_defaults_to_single_worker'`, and all selected tests passed. 
- Running the entire `python -m pytest -q tests/test_relay.py` in this environment showed unrelated pre-existing failures in other API v1 relay client tests, so the change is validated only by the focused tests above. 
- `pre-commit run --all-files` completed successfully in this environment. 
- Attempted `docker build -t tokenplace-relay:test .` but the execution environment lacks Docker (`docker: command not found`), so the smoke container validation could not be performed here.

------
[Codex Task](https://chatgpt.com/codex/cloud/tasks/task_e_6a129ae7cee8832fbbb1b1bfbdbbbce3)


futuroptimist:

PR #949, opened 2026-05-24
Make relay health safe for single-replica Sugarkube and default to one worker

futuroptimist:
### Motivation
- Prepare the relay runtime for Sugarkube phase‑1 single-replica, relay-only deployments where no in-cluster GPU/backend is required.
- Keep readiness green by default when upstream/GPU services are absent while allowing an opt-in strict gate for deployments that need it.
- Ensure the container entrypoint is POSIX-shell safe and the default Gunicorn worker count does not split in-memory state.

### Description
- Update `docker/relay/entrypoint.sh` to use a POSIX-safe shell invocation and flags (`#!/usr/bin/env sh` + `set -eu`) and change the default Gunicorn worker count to one while preserving `RELAY_WORKERS` override and all existing runtime knobs (bind, threads, timeouts, worker tmp dir, logs). (`docker/relay/entrypoint.sh`)
- Add a boolean env helper `_env_truthy` and new env var `TOKENPLACE_RELAY_REQUIRE_UPSTREAM_HEALTH` (default `false`) and wire it into app config at startup. (`relay.py`)
- Modify `/healthz` readiness logic so unresolved upstream/GPU DNS only causes `degraded`/`503` when `require_upstream_health` is enabled; preserve draining behavior (`503` + `Retry-After: 0`) and leave `/livez` unchanged. (`relay.py`)
- Add focused tests that assert default relay-only `/healthz` stays `ok` with an unresolvable upstream, that `TOKENPLACE_RELAY_REQUIRE_UPSTREAM_HEALTH=1` restores degraded/503 behavior, and that the entrypoint script defaults to one worker. (`tests/test_relay.py`)

### Testing
- Ran focused tests: `python -m pytest -q tests/test_relay.py -k 'healthz_relay_only_default or healthz_can_enforce_unresolvable_upstream_failure or relay_entrypoint_defaults_to_one_worker'` — all selected tests passed.
- Ran the full relay test file: `python -m pytest -q tests/test_relay.py` — test run completed but included 6 failing tests unrelated to the health/entrypoint changes (pre-existing API v1 relay-client and poll-wait expectations); the new tests still pass in isolation.
- Ran `pre-commit run --all-files` which passed in this environment.
- Attempted `docker build -t tokenplace-relay:test .` but the `docker` CLI is unavailable in this runner, so image build/ smoke-run could not be verified here.

------
[Codex Task](https://chatgpt.com/codex/cloud/tasks/task_e_6a129ae7cee8832fbbb1b1bfbdbbbce3)


futuroptimist:

PR #975, opened 2026-05-24
Fix macOS desktop bridge startup failure by bundling minimal `requests` shim

futuroptimist:
### Motivation
- Packaged macOS Tauri desktop operator could exit early with `No module named 'requests'` because the compute-node bridge import path relied on a third-party `requests` package that may not be present in packaged/clean runtimes.
- The fix must make packaged desktop startup deterministic without relying on users' global Python, Homebrew, or contributor virtualenvs.

### Description
- Add a minimal, urllib-backed `requests` compatibility shim at `desktop-tauri/src-tauri/python/requests.py` implementing `get`, `post`, `Response`, and the common `requests` exception types used by the bridge and relay client.
- Update the packaged-layout e2e probe `desktop-tauri/scripts/test_packaged_operator_e2e.py` to copy the shim into the packaged resources and force packaged-resource-first import resolution by setting `PYTHONPATH` to the packaged `resources/python` and `PYTHONNOUSERSITE=1` so local/global site-packages cannot mask undeclared dependencies.
- Add outage documentation `outages/2026-05-24-macos-desktop-bridge-requests-dependency.md` describing the incident, root cause, fix, and follow-ups.
- Files changed: `desktop-tauri/src-tauri/python/requests.py`, `desktop-tauri/scripts/test_packaged_operator_e2e.py`, and `outages/2026-05-24-macos-desktop-bridge-requests-dependency.md`.

### Testing
- Ran the packaged-operator probe with `python desktop-tauri/scripts/test_packaged_operator_e2e.py`, which completed successfully in this environment and validated the bridge startup/inspect path when packaged resources are used (✅).
- Ran `pre-commit run --all-files`, which executed and passed most hooks but failed on unrelated YAML parsing in `charts/tokenplace/templates/*.yaml` (existing repository issue) and therefore did not block this change (⚠️).
- Verification summary: the bridge startup path no longer fails with `No module named 'requests'` when executed from the packaged-layout probe because the shim is present and `PYTHONPATH` prioritizes packaged resources, and the e2e probe asserts the bridge emits `started` and `registered` events.

Notes and limitations: this change implements the smallest targeted fix by shipping a tiny compatibility shim rather than adding a global dependency; full macOS packaged-app GUI e2e in CI is still environment-dependent and out of scope for this patch, so follow-up should enable macOS-packaged smoke tests in CI if practical.

------
[Codex Task](https://chatgpt.com/codex/cloud/tasks/task_e_6a138364cb3c832f8adeec656d911d81)

### DIFF
--- a/desktop-tauri/scripts/test_packaged_operator_e2e.py
+++ b/desktop-tauri/scripts/test_packaged_operator_e2e.py
@@ -58,6 +58,7 @@ def create_packaged_layout(tmp_root: Path) -> Path:
         "inference_sidecar.py",
         "model_bridge.py",
         "path_bootstrap.py",
+        "requests.py",
     ):
         shutil.copy2(
             REPO_ROOT / "desktop-tauri" / "src-tauri" / "python" / filename,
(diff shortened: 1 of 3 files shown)


futuroptimist:

PR #978, opened 2026-05-25
Preflight desktop Python deps and add no-relay-autostart guards

futuroptimist:
### Motivation
- The desktop compute-node bridge was exiting early on macOS with `No module named 'psutil'` because `psutil` was imported at module-import time and the packaged-like interpreter (launched with `PYTHONNOUSERSITE=1`) might not have the dependency installed.  The bridge must fail with an actionable diagnostic or proactively provision dependencies before the operator runtime path proceeds. 
- The desktop Tauri app must never implicitly start or bundle `relay.py` or bind to `localhost:5010`; static/lifecycle guards were missing so a phantom `relay.py` could persist after app shutdown.
- Capture the incident and the corrective changes in an `outages/` entry to document root cause, why tests missed it, and the regression coverage added.

### Description
- Add `ensure_desktop_python_dependencies(required_modules: list[str])` to `desktop-tauri/src-tauri/python/desktop_runtime_setup.py` which probes imports and returns structured diagnostics including `missing`, `interpreter`, and `prefix`.
- Add a dependency preflight in `desktop-tauri/src-tauri/python/compute_node_bridge.py` that runs before runtime imports; if required modules are missing the bridge emits a structured `status` with an actionable `last_error` and exits early to avoid the cryptic startup failure.
- Add a static unit test `tests/unit/test_desktop_no_relay_autostart.py` that asserts the Tauri bundle config and desktop startup sources do not include `relay.py` or `:5010` markers to prevent accidental bundling/auto-start of the relay surface.
- Add outage documentation `outages/2026-05-25-desktop-macos-bridge-deps-and-relay-autostart-guard.md` and matching JSON summarizing symptoms, root causes, fix, and test commands.

### Testing
- Ran unit tests: `python -m pytest tests/unit/test_desktop_compute_node_bridge.py tests/unit/test_desktop_runtime_setup.py tests/unit/test_resource_monitor.py tests/unit/test_desktop_no_relay_autostart.py` and all tests passed (108 passed).
- Ran repository checks: `pre-commit run --all-files` — hook run succeeded for the changed checks, but overall pre-commit failed due to existing YAML parse errors in unrelated Helm chart templates (`charts/tokenplace/templates/*.yaml`); those failures are pre-existing and outside this change.
- Local commit created with the changes and added outage artifacts; the change surface was kept minimal and covered by unit tests added/updated above.

------
[Codex Task](https://chatgpt.com/codex/cloud/tasks/task_e_6a1493d11654832fa8306483833eca0a)

### DIFF
--- a/desktop-tauri/src-tauri/python/compute_node_bridge.py
+++ b/desktop-tauri/src-tauri/python/compute_node_bridge.py
@@ -26,6 +26,7 @@ try:
     from desktop_runtime_setup import (
         desktop_gpu_runtime_failure_message,
         ensure_desktop_llama_runtime,
+        ensure_desktop_python_dependencies,
         maybe_reexec_for_runtime_refresh,
     )
 except ModuleNotFoundError:
@@ -39,6 +40,9 @@ except ModuleNotFoundError:
             "runtime_action": "unavailable",
             "fallback_reason": "desktop_runtime_setup module missing",
         }
+
+    def ensure_desktop_python_dependencies(_required_modules: list[str]) -> Dict[str, str]:
+        return {"ok": "1", "missing": ""}
 
     def maybe_reexec_for_runtime_refresh(
         _runtime_setup: Dict[str, str], *, allow_reexec: bool = True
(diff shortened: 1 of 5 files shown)


futuroptimist:

PR #1014, opened 2026-05-27
Clarify mandatory E2EE messaging on relay landing page and add guard test

futuroptimist:
### Motivation
- The landing-page wording implied end-to-end encryption could be optional, which conflicts with the repo invariant that relay paths must remain ciphertext-only and fail closed for plaintext relay dispatch. 
- Make the contract explicit in user-facing docs and add a regression guard so the messaging cannot be accidentally reverted.

### Description
- Update `static/index.html` to reword the “Encrypted API Usage” section to `End-to-End Encryption (Always On)` and state that E2EE is mandatory across client↔relay↔server paths. 
- Tweak the legacy `/faucet` docstring in `relay.py` to clarify the relay carries ciphertext envelopes and routing metadata rather than plaintext. 
- Add `tests/unit/test_relay_index_e2ee_copy.py` to assert the landing-page contains the required mandatory-E2EE phrasing and that the old optional-privacy copy was removed.

### Testing
- Ran `pytest -q tests/unit/test_relay_index_e2ee_copy.py tests/unit/test_e2ee_relay_invariant.py`, which passed (`7 passed`). 
- Ran `pre-commit run --all-files`, which failed on existing unrelated YAML parse errors in Helm templates under `charts/tokenplace/templates/*.yaml` (these failures are pre-existing and not caused by this change). 
- Existing E2EE invariant tests were exercised and continue to enforce that plaintext is not stored, routed, or logged by relay paths.

------
[Codex Task](https://chatgpt.com/codex/cloud/tasks/task_e_6a16a4b3f2a4832fa70e87a09a1e5ede)

### DIFF
--- a/relay.py
+++ b/relay.py
@@ -1171,9 +1171,9 @@ def faucet():
     with a list of chat messages between the user and the assistant. Clients and servers may implement
     diffing mechanisms to determine if conversations were altered, but this is intentionally built in a
     stateless manner to minimize complexity and maximize scalability. It's trivially easy to identify
-    tampering from either party, so enforcing it on the relay is a non-goal, especially since communication
-    between the client and the server is intended to be end-to-end-encrypted. Servers and clients can choose
-    how to handle this as they see fit.
+    tampering from either party, so enforcing it on the relay is a non-goal. Communication between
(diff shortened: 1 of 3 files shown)


futuroptimist:

PR #1049, opened 2026-05-29
Keep desktop API v1 relay registration alive after no-work polls

futuroptimist:
### Motivation

- Staging showed desktop nodes registering and receiving a first no-work poll but then aging out before a later browser/API v1 request, causing `/api/v1/relay/servers/next` to return 503.  
- The root cause was the bridge treating a single successful register/no-work poll as sufficient and sleeping long enough that the relay lease could expire during model warmup or other local pauses.  
- This change preserves API v1 E2EE invariants and keeps the bridge/relay interactions API v1-only while improving registration freshness and diagnostics.

### Description

- Track per-relay registration lease metadata and short public-key fingerprints in the client and store heartbeat timing so the client can reason about lease freshness (`utils/networking/relay_client.py`).  
- Re-register proactively when the client detects unknown-node (404), public-key changes, or local lease-expiry risk and record reasons using safe fingerprints (no raw full public keys in new logs).  
- Annotate poll responses with `lease_seconds` and `registration_fresh_for_seconds`, clamp local sleep cadence below the relay lease, and use those signals to make `Registered: yes` reflect a recently confirmed heartbeat rather than a stale local flag (`desktop-tauri/src-tauri/python/compute_node_bridge.py`).  
- Run post-registration warmup in a background thread so warmup does not block the poll/heartbeat cadence, and add targeted tests covering lease-risk reregistration, no-work poll freshness, and stale-registration detection (unit + integration tests updated/added).

### Testing

- Ran unit and integration suites: `pytest tests/unit/test_relay_client.py` (118 passed), `pytest tests/unit/test_desktop_compute_node_bridge.py` (39 passed), `pytest tests/integration/test_api_v1_desktop_bridge_e2ee.py` (7 passed), and `pytest tests/test_relay.py -k "api_v1 or relay or lease or heartbeat"` (77 passed).  
- Combined/targeted runs also reported all relevant tests passing (total reported run summary: 202/202 items in combined suites, all passed for the API v1/relay/lease/heartbeat scenarios).  
- Ran `pre-commit run --all-files`; hooks completed but the `check-yaml` hook reported failures on existing Helm template YAML files under `charts/tokenplace/templates/*.yaml` (these are pre-existing template parsing issues and not caused by this change); other pre-commit hooks (formatting, codespell, bandit, vulture) passed and `end-of-file-fixer` normalized a missing final newline in `docs/relay_sugarkube_onboarding.md`.

------
[Codex Task](https://chatgpt.com/codex/cloud/tasks/task_e_6a18e56dd160832fb7eb50847933e64a)


futuroptimist:

PR #1147, opened 2026-06-08
Populate landing chat model selector from GET /api/v1/models and use selection for chat

futuroptimist:
### Motivation
- Surface the API v1 model list on the landing page and let users choose which model to send chat requests to, while keeping API v1 as the single non-streaming launch surface for v0.1.x.
- Keep the landing chat resilient when the model list fails and avoid hard-coding a model id in the client request payload.

### Description
- Added model-list state and fetching to the Vue landing app in `static/chat.js` (`modelEntries`, `selectedModelId`, `isLoadingModels`, `modelListError`) and a computed `canSendMessage` gate that requires client/server keys, a selected model, and no in-flight response.
- Implemented `fetchModels()` which calls `GET /api/v1/models`, populates the dropdown entries, selects the first model by default, and falls back to a packaged emergency model (`EMERGENCY_FALLBACK_MODEL_ID = 'llama-3-8b-instruct'`) with a non-blocking error message if listing fails.
- Replaced the hard-coded model id in chat requests with the selected model (`model: this.selectedModelId`) while preserving encrypted API v1 envelope behavior and non-streaming flow.
- Added minimal UI in `static/index.html` (label, `<select>` with `data-testid=

------
[Codex Task](https://chatgpt.com/codex/cloud/tasks/task_e_6a260310b190832fa71a4f2f193f7adc)

### DIFF
--- a/static/chat.js
+++ b/static/chat.js
@@ -1,5 +1,6 @@
 const ASSISTANT_GENERIC_FALLBACK_MESSAGE = 'Sorry, I encountered an issue generating a response. Please try again.';
 const ASSISTANT_INVALID_RELAY_RESPONSE_MESSAGE = 'Sorry, the relay returned an invalid response. Please try again.';
+const EMERGENCY_FALLBACK_MODEL_ID = 'llama-3-8b-instruct';
 
 new Vue({
     el: '#app',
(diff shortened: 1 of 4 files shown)


futuroptimist:

PR #1244, opened 2026-06-20
Route relay-only API v1 chat through compute nodes

futuroptimist:
### Motivation

- Relay-only deployments that publish API v1 via a public/internal relay URL must route chat requests to registered compute nodes and must not require local `llama-cpp-python` to be installed, nor surface the local dependency error when relay dispatch is expected.

### Description

- Force distributed provider selection when the resolved distributed target is marked `relay_only` by changing `_read_api_v1_provider_env` to set `mode="distributed"` and `distributed_fallback_enabled=False` so relay-only public/internal relay targets are used before any local inference path.
- Preserve existing behavior for explicit local deployments by leaving non-relay and explicit `local` modes unchanged.
- Add regression tests in `tests/unit/test_api_v1_launch_contract.py` that verify (1) relay-only chat exercises the distributed compute-node path and does not call local llama inference, (2) relay-only with no registered compute node returns a relay-specific 503 (no `llama-cpp-python` message) and preserves public API v1 CORS, and (3) explicit local mode still reports missing `llama-cpp-python` when local inference is selected.

### Testing

- Ran `pytest -q tests/unit/test_api_v1_launch_contract.py` and the targeted launch-contract tests passed.
- Ran `pytest -q tests/unit/test_api_v1_launch_contract.py tests/unit/test_api_v1_cors.py tests/test_relay.py` and the combined suite passed.
- Ran the full unit suite with `pytest -q tests/unit` and it passed (all unit tests passed in this environment).
- Verified CORS and error-string search with `rg` and ran `pre-commit run --all-files`, both of which completed successfully.

------
[Codex Task](https://chatgpt.com/codex/cloud/tasks/task_e_6a35ffc88e38832f89b1237841b6c485)

### DIFF
--- a/api/v1/compute_provider.py
+++ b/api/v1/compute_provider.py
@@ -730,14 +730,27 @@ def get_api_v1_compute_provider() -> ApiV1ComputeProvider:
 
 
 def _read_api_v1_provider_env() -> tuple[str, DistributedTargetSelection, bool]:
-    """Read and normalize API v1 provider environment configuration."""
+    """Read and normalize API v1 provider environment configuration.
+
+    Relay-only deployments publish API v1 from ``relay.py`` and rely on
+    registered compute nodes instead of an in-process llama.cpp runtime.  When
+    a relay-only public/internal relay URL is configured, select the
+    distributed provider before local inference can attempt to import or load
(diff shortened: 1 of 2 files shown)


futuroptimist:

PR #1245, opened 2026-06-20
Route relay-only API v1 chat through compute nodes

futuroptimist:
### Motivation
- In relay-only deployments the public API v1 `/chat/completions` must not attempt to load local `llama-cpp-python` and must route to registered compute nodes instead of failing with a local-dependency message.
- When no compute node is available the API should return an explicit relay-specific error (e.g. `no_registered_compute_nodes`) that operators can act on, while preserving existing local-inference behavior for non-relay deployments.

### Description
- Added relay-only provider selection helpers and heuristics in `api/v1/routes.py`: `_env_truthy`, `_has_explicit_relay_upstream_config`, `_request_relay_only_target_selection`, and `_select_chat_completion_provider` to prefer a distributed (relay) compute-node path in relay-only mode.
- Replaced the inline provider selection in the chat handler with ` _select_chat_completion_provider(...)` so relay-only routing is evaluated before any local model/loading paths are considered.
- Preserved forced desktop-bridge distributed routing and the existing local fallback behavior when relay-only heuristics do not apply.
- Added focused tests in `tests/unit/test_api_v1_routes_additional.py` covering: relay-only selects the distributed compute-node path without calling local llama, and relay-only with no compute node returns the relay-specific `no_registered_compute_nodes` error and preserves public CORS headers.

### Testing
- Ran the focused unit tests with `pytest -q tests/unit/test_api_v1_routes_additional.py tests/unit/test_api_v1_cors.py` and they passed.
- Ran the broader suites `pytest -q tests/test_relay.py tests/unit/test_api_v1_launch_contract.py tests/unit/test_api_v1_cors.py tests/unit` and observed `170 passed` in the run recorded in the transcript.
- Ran repository checks `pre-commit run --all-files` which completed successfully, and an `rg` search for deployed/error strings to verify the new error semantics was executed.
- New tests added: `test_chat_completion_relay_only_uses_compute_node_path_without_local_llama` and `test_chat_completion_relay_only_no_compute_node_returns_relay_specific_error`, both passing in CI-local runs; residual risk: the relay-only heuristics use environment/config heuristics (public base URL and upstream flags), so monitor staging to ensure the selection matches actual deployment environment in production promotion.

------
[Codex Task](https://chatgpt.com/codex/cloud/tasks/task_e_6a35ffe0e7e4832f9b02f997a8e712c8)

### DIFF
--- a/api/v1/routes.py
+++ b/api/v1/routes.py
@@ -5,7 +5,7 @@ This module follows OpenAI API conventions to serve as a drop-in replacement.
 
 from __future__ import annotations
 
-from flask import Blueprint, request, jsonify
+from flask import Blueprint, current_app, request, jsonify
 import base64
 import time
 import zlib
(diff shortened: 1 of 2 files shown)


futuroptimist:

PR #1246, opened 2026-06-20
Route API v1 chat to distributed compute in relay‑only deployments

futuroptimist:
### Motivation
- Relay-only staging was incorrectly selecting the local inference path and failing with "llama-cpp-python is not installed" instead of dispatching to registered compute nodes or returning a relay-specific error. 
- Public API v1 chat must not require local `llama-cpp-python` when the process is acting as a relay-only deployment, and errors must be actionable and distinguish relay availability from local dependency failures.

### Description
- Add relay-only detection and provider selection in `api/v1/routes.py` via `_current_app_is_relay_only()` and `_select_chat_completion_provider()` so the chat handler prefers the `distributed` provider with `distributed_fallback_enabled=False` when the Flask app is relay-only. 
- Use configured `public_base_url` for relay-only dispatch with `_relay_only_target_selection()` so relay-only public relays route to their own compute target. 
- Replace the inline provider selection in the chat handler with the new `_select_chat_completion_provider()` to avoid falling through to local inference when relay-only. 
- Add regression tests in `tests/unit/test_api_v1_launch_contract.py` to assert that relay-only chat chooses a distributed provider (not local), returns an OpenAI-compatible completion on success, and returns a relay-specific 503 with wildcard CORS and an explicit `no_registered_compute_nodes` error when no compute node is available.

### Testing
- Ran focused unit and integration suites: `pytest -q tests/test_relay.py` (still passing), `pytest -q tests/unit/test_api_v1_launch_contract.py` (passed), `pytest -q tests/unit/test_api_v1_cors.py` (passed), and `pytest -q tests/unit` (all unit tests passed). 
- Verified the staging relay regression: `pytest -q tests/test_api.py::test_api_v1_chat_completion_staging_no_nodes_returns_clear_503` passed and the new relay-specific test `tests/unit/test_api_v1_launch_contract.py::test_relay_only_chat_no_compute_node_reports_relay_specific_error_with_cors` passed. 
- Ran repository checks via `pre-commit run --all-files` which executed project checks and the test runner hooks; the run completed successfully.

------
[Codex Task](https://chatgpt.com/codex/cloud/tasks/task_e_6a35ffd586e4832f9ce70abe26f110f3)

### DIFF
--- a/api/v1/routes.py
+++ b/api/v1/routes.py
@@ -5,7 +5,7 @@ This module follows OpenAI API conventions to serve as a drop-in replacement.
 
 from __future__ import annotations
 
-from flask import Blueprint, request, jsonify
+from flask import Blueprint, current_app, request, jsonify
 import base64
 import time
 import zlib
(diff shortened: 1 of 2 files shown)


futuroptimist:

PR #1314, opened 2026-06-24
Fix desktop bridge startup diagnostics for missing context_profiles

futuroptimist:
### Motivation
- Packaged desktop operator startup could exit before emitting a startup event when `utils.context_profiles` failed to import in release-like resource layouts, leaving the UI with only the generic "compute-node bridge exited before emitting a startup event" message.

### Description
- Defer and catch `utils.context_profiles` import errors in `desktop-tauri/src-tauri/python/compute_node_bridge.py` and record the exception so the bridge can emit a structured `error` event instead of crashing. 
- Add `_context_profiles_unavailable_message`, `_safe_import_root_for_diagnostics`, and `_normalize_context_tier_for_bridge` helpers to produce a sanitized diagnostic string that includes safe fields like `interpreter` and `import_root` but not secrets or model paths. 
- Normalize the requested context tier before importing the compute runtime and fail early with a structured startup `error` if context profile resources are unavailable. 
- Remove `model_path` from early startup `error` payloads to avoid exposing model paths on the earliest failures. 
- Add a release-like subprocess regression test in `tests/unit/test_desktop_compute_node_bridge.py` that simulates missing bundled `utils.context_profiles` and asserts the bridge emits a safe structured error; adjust an existing test to reflect `model_path` omission.

### Testing
- Ran `python -m pytest tests/unit/test_context_profiles.py -q` and it passed (3 tests). 
- Ran `python -m pytest tests/unit/test_desktop_compute_node_bridge.py -q` and it passed (105 tests). 
- Ran desktop front-end checks with `cd desktop-tauri && npm test -- --run` and `npm run build`, both succeeded. 
- Attempted `cd desktop-tauri/src-tauri && cargo test` but the run is blocked by missing system dependency (`glib-2.0.pc` / `pkg-config`); cargo reported missing `glib-2.0` (environment issue). 
- Ran `pre-commit run --all-files` and `git diff --check`, both completed successfully. 
- Residual risk: packaged / installer layouts should be validated on actual platform artifacts (macOS/Windows packaged bundles) and a follow-up manual verification is recommended to ensure the bundled `utils` path semantics match installer resource layout.

------
[Codex Task](https://chatgpt.com/codex/cloud/tasks/task_e_6a3c41cacf34832f9ac67105a4be842a)

### DIFF
--- a/desktop-tauri/src-tauri/python/compute_node_bridge.py
+++ b/desktop-tauri/src-tauri/python/compute_node_bridge.py
@@ -27,7 +27,14 @@ from path_bootstrap import ensure_runtime_import_paths
 
 ensure_runtime_import_paths(__file__, avoid_llama_cpp_shadowing=True)
 from pathlib import Path
-from utils.context_profiles import apply_context_profile, get_context_profile
+
+_CONTEXT_PROFILES_IMPORT_ERROR: Optional[BaseException] = None
+try:
+    from utils.context_profiles import apply_context_profile, normalize_context_tier
+except Exception as exc:  # pragma: no cover - exercised by subprocess startup tests
+    apply_context_profile = None  # type: ignore[assignment]
+    normalize_context_tier = None  # type: ignore[assignment]
+    _CONTEXT_PROFILES_IMPORT_ERROR = exc
 
 try:
     from desktop_runtime_setup import (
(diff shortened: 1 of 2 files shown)


futuroptimist:

PR #1385, opened 2026-07-03
Probe child subprocess for YaRN/RoPE support and add packaged-desktop Qwen64k e2e test

futuroptimist:
### Motivation
- Packaged macOS desktop runs could incorrectly decide Qwen 64K YaRN/RoPE constructor kwargs are unsupported because the parent process only saw a subprocess facade/proxy signature, causing false-negative startup failures. 
- The runtime capability detection must run inside the real child subprocess (the packaged runtime) to avoid trusting the parent facade and to produce actionable diagnostics for repair.

### Description
- Move and extend the subprocess capability probe to return a safe JSON capability payload from the real child `llama_cpp` import that includes constructor signature classification, `**kwargs` presence, per-kwarg support/status, yarn enum discovery, and other diagnostics used by the manager. (changes in `utils/llm/model_manager.py`).
- Treat `**kwargs` / signature-unavailable results as `unknown`/capable rather than automatically marking all explicit kwargs unsupported, and only use the numeric YaRN fallback when the real child probe indicates `rope_scaling_type` is accepted. (changes in `utils/llm/model_manager.py`).
- Make the subprocess facade (`_SubprocessLlamaCppModule`) attempt a child probe when no explicit desktop probe is provided, and propagate probe-derived fields (constructor support/status, yarn enum value/source, module path/version, etc.) into the facade payload used by parent-side logic. (changes in `utils/llm/model_manager.py`).
- Update the Qwen 64K init path to rely on child-probe diagnostics before failing and to include richer, safe diagnostics (child signature classification, yarn resolver source, constructor kwargs attempted, parent facade type). (changes in `utils/llm/model_manager.py`).
- Add packaged-desktop regression/e2e tests that exercise the subprocess facade path with a fake child `llama_cpp` module that either supports or lacks YaRN kwargs, asserting that: explicit YaRN kwargs are passed to the child (`n_ctx=65536`, `rope_scaling_type`, `yarn_ext_factor=2.0`, `yarn_orig_ctx=32768`) when supported and that startup fails closed with precise diagnostics when unsupported. (new `tests/integration/test_desktop_qwen64k_packaged_runtime.py`).
- Adjust a unit test to explicitly construct a subprocess facade with supplied `desktop_runtime_probe` attributes so the memory-profile helpers do not trust proxy signatures alone. (tests updated: `tests/unit/test_model_manager.py`).

### Testing
- Ran the focused test suites and the new packaged-desktop e2e: `python -m pytest tests/unit/test_model_manager.py tests/integration/test_desktop_qwen64k_packaged_runtime.py -q`, and additionally exercised the related unit/integration suites described in the task. All executed tests passed locally. 
- Key commands run: `python -m pytest tests/unit/test_compute_node_runtime.py -q`, `python -m pytest tests/unit/test_context_profiles.py -q`, `python -m pytest tests/unit/test_desktop_runtime_setup.py -q`, `python -m pytest tests/unit/test_desktop_gpu_packaging.py -q`, `python -m pytest tests/integration/test_api_v1_desktop_bridge_e2ee.py -q`, and `python -m pytest tests/integration/test_desktop_qwen64k_packaged_runtime.py -q`, all of which succeeded. 
- Pre-commit and project checks were run via `pre-commit run --all-files` and returned clean results.

------
[Codex Task](https://chatgpt.com/codex/cloud/tasks/task_e_6a47ed81d2fc832f9a7aa1b5de438572)

### DIFF
--- a/tests/integration/test_desktop_qwen64k_packaged_runtime.py
+++ b/tests/integration/test_desktop_qwen64k_packaged_runtime.py
@@ -1,0 +1,112 @@
+import json
+import sys
+from pathlib import Path
+from unittest.mock import MagicMock, patch
+
+from utils.context_profiles import apply_context_profile
+from utils.llm import model_manager as model_manager_module
+from utils.llm.model_manager import ModelManager
+
+
+def _config(tmp_path):
+    values = {
+        'model.profile_id': 'qwen3-8b-q4-k-m',
+        'model.context_size': 8192,
+        'model.use_mock': False,
+        'model.n_gpu_layers': 0,
+        'model.enforce_gpu_memory_headroom': False,
+        'paths.models_dir': str(tmp_path),
+    }
+    config = MagicMock(is_production=False)
+    config.get.side_effect = lambda key, default=None: values.get(key, default)
+    config.set.side_effect = lambda key, value: values.__setitem__(key, value)
+    return config
+
+
+def _write_fake_llama_cpp(site_dir: Path, record_path: Path, *, supports_yarn: bool) -> None:
+    pkg = site_dir / 'llama_cpp'
+    pkg.mkdir(parents=True)
+    if supports_yarn:
+        init = f'''
+__version__ = '0.3.32'
+LLAMA_ROPE_SCALING_TYPE_YARN = 2
+GGML_USE_METAL = True
+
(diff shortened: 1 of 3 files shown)


futuroptimist:

PR #1404, opened 2026-07-05
Use rendered GGUF/Jinja prompt + plain completion for Qwen API v1 (render-then-complete)

futuroptimist:
### Motivation
- The Qwen64K readiness and first-generation smoke were failing inside the high-level `create_chat_completion` path in the packaged subprocess, so generation must use the already-verified GGUF/Jinja render/tokenize path and call plain completion in the child worker. 
- Preserve the existing non-thinking Qwen controls, exact context admission token accounting, and relay E2EE while avoiding returning rendered prompts to the parent.

### Description
- Add a subprocess worker bridge `create_completion_from_rendered_prompt` that renders messages with the GGUF/Jinja template inside the child, calls the runtime plain `create_completion(prompt=..., ...)` with a conservative allowlist of kwargs, and returns only the runtime result/demarcated safe diagnostics. (changes in `utils/llm/model_manager.py` and its subprocess handler.)
- Add RelayClient support for a render-then-complete generation path with kwarg filtering and stop-sequence handling via `_api_v1_create_completion_from_rendered_prompt_filtered`, and route Qwen API v1 non-thinking generation to this path when available while keeping the old chat path for non-Qwen or debug cases. (changes in `utils/networking/relay_client.py`.)
- Preserve and extend safety/normalization: strip trailing Qwen end markers (`<|im_end|>`), retain the empty-`<think></think>` normalizer and reject leaked `<think>` reasoning, and map precise render-complete failure reasons to distinct safe diagnostics. (changes in `utils/networking/relay_client.py` and `utils/compute_node_runtime.py`.)
- Update the packaged Qwen 64K fake-runtime integration tests to assert the new render-then-complete flow, filter unsupported plain-completion kwargs (e.g. `top_k`), and verify rendered prompt never leaves diagnostics; adapt tests expecting `create_chat_completion` to use the render-complete bridge where appropriate. (changes in `tests/integration/test_desktop_qwen64k_packaged_runtime.py`.)

### Testing
- Ran focused test suite: `python -m pytest tests/unit/test_model_manager.py -q`, `python -m pytest tests/unit/test_relay_client.py -q`, `python -m pytest tests/unit/test_compute_node_runtime.py -q`, `python -m pytest tests/unit/test_context_profiles.py -q`, `python -m pytest tests/integration/test_api_v1_desktop_bridge_e2ee.py -q`, and `python -m pytest tests/integration/test_desktop_qwen64k_packaged_runtime.py -q`, and all targeted tests passed (unit and integration suites reported success). 
- Ran the full combined test command used in verification and the pre-commit hooks: `pre-commit run --all-files` and `git diff --check`, and they completed with no failing hooks or check errors.
- The changes include a new packaged-runtime-focused regression test that models the original failure mode (child `create_chat_completion` raises while `create_completion` succeeds) and this test now passes, preventing the previous readiness failure on Qwen 64K.

------
[Codex Task](https://chatgpt.com/codex/cloud/tasks/task_e_6a49de441b84832f8fe78c9487d9ab79)

### DIFF
--- a/tests/integration/test_desktop_qwen64k_packaged_runtime.py
+++ b/tests/integration/test_desktop_qwen64k_packaged_runtime.py
@@ -61,6 +61,8 @@ class Llama:
     def tokenize(self, payload, add_bos=False):
         return [1] * 42
     def create_chat_completion(self, **kwargs):
+        raise RuntimeError('high-level chat path disabled for Qwen test')
+    def create_completion(self, prompt, **kwargs):
 """
         + completion_branch,
         encoding='utf-8',
@@ -153,7 +155,11 @@ class _Qwen64kFakeRuntime:
         return "<redacted-test-template>"
 
     def create_chat_completion(self, **_kwargs):
-        return {"choices": [{"message": {"role": "assistant", "content": "ok"}}]}
+        raise AssertionError("Qwen API v1 must not use create_chat_completion")
+
+    def create_completion_from_rendered_prompt(self, **kwargs):
+        self.last_render_complete_kwargs = kwargs
+        return {"choices": [{"text": "ok"}]}
 
 
 def _model_manager(runtime):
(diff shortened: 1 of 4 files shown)


futuroptimist:

PR #1413, opened 2026-07-07
Harden Qwen API v1 non-thinking readiness and remove automatic /no_think injection

futuroptimist:
### Motivation
- Remove hidden `/no_think` message mutation and enforce Qwen API v1 non-thinking via hard render/template controls and runtime flags. 
- Fail closed and surface safe scalar diagnostics when a packaged runtime rejects required non-thinking or other bounded generation kwargs. 
- Ensure plain-completion paths remain bounded and that `enable_thinking=False` is never dropped or retried away.

### Description
- Removed automatic `/no_think` injection: `_api_v1_qwen_no_think_messages()` is now a no-op compatibility shim that returns copies of messages without mutating user content and `_API_V1_QWEN_NON_THINKING_POLICY` no longer contains `message_control`.
- Added deterministic Qwen non-thinking ChatML fallback: `_render_qwen_api_v1_non_thinking_template(messages, llama, *, add_generation_prompt=True)` renders messages with Qwen ChatML markers and appends an explicit empty `<think></think>` assistant prefix when requested, failing closed on malformed roles/content and never injecting `/no_think`.
- Hardened runtime template selection in `_render_chat_with_runtime_template`: require/validate `enable_thinking=False` semantics, prefer direct `apply_chat_template` only if it accepts the hard switch, attempt GGUF/Jinja if available, and fall back to the trusted Qwen template only when metadata/provider evidence indicates Qwen and the other paths cannot prove they honored `enable_thinking=False`.
- Protected `enable_thinking` at parent relay filter in `_api_v1_create_completion_from_rendered_prompt_filtered()`: always set `completion_kwargs['enable_thinking']=False` for Qwen, add it to the never-filter set, and fail closed with safe diagnostics if a child rejects `enable_thinking`.
- Kept plain-completion bounded and added capability diagnostics: added `plain_completion_*` capability fields, recorded attempted plain-completion methods and attempted generation kwargs, and ensured no unbounded fallback is attempted when `max_tokens` is rejected.
- Strengthened safe diagnostic allowlists and readiness mapping: added new safe keys and mapped render-vs-plain unexpected-kwarg errors to distinct readiness reasons so UI status reports include specific safe scalars like `rejected_generation_kwarg`, `method`, attempted kwargs/methods, and capability booleans.
- Tests updated to reflect removal of automatic `/no_think` (preserve literal user `/no_think`), to assert `enable_thinking=False` is always sent and never retried away, and to validate the Qwen fallback template output and bounded plain-completion behavior.

### Testing
- Ran targeted unit and integration suites locally; all modified and relevant test suites passed after the change: 
  - `python -m pytest tests/unit/test_model_manager.py -q` — passed
  - `python -m pytest tests/unit/test_relay_client.py -q` — passed
  - `python -m pytest tests/unit/test_compute_node_runtime.py -q` — passed
  - `python -m pytest tests/unit/test_desktop_compute_node_bridge.py -q` — passed (focused rerun fixed a transient)
  - `python -m pytest tests/integration/test_api_v1_desktop_bridge_e2ee.py -q` — passed
  - `python -m pytest tests/integration/test_desktop_qwen64k_packaged_runtime.py -q` — passed
  - `python -m pytest tests/unit -q` — passed (full-unit run: 1977 passed, 1 skipped)
  - `git diff --check` — no whitespace/errors
  - `pre-commit run --all-files` — passed
- Notes about remaining deployed diagnosis: the exact rejected kwarg/method causing the previously observed deployed readiness collapse could not be reproduced locally because the real packaged Metal Qwen runtime is not available; the change adds precise safe diagnostics (`rejected_generation_kwarg`, `method`, `attempted_generation_kwargs`, `attempted_plain_completion_methods`, `plain_completion_*` capability booleans) so the next deployed readiness failure should reveal the concrete rejected kwarg/method without exposing plaintext model or prompt data.

------
[Codex Task](https://chatgpt.com/codex/cloud/tasks/task_e_6a4c96b7a000832f963b10080b2b772f)

### DIFF
--- a/tests/unit/test_compute_node_runtime.py
+++ b/tests/unit/test_compute_node_runtime.py
@@ -747,7 +747,7 @@ def test_compute_node_runtime_readiness_smoke_completion_passes(monkeypatch):
     assert llm_runtime.completion_kwargs["max_tokens"] == 64
     assert "stream" not in llm_runtime.completion_kwargs
     assert "stop" not in llm_runtime.completion_kwargs
-    assert llm_runtime.completion_kwargs["messages"][-1]["content"].startswith("/no_think")
+    assert not llm_runtime.completion_kwargs["messages"][-1]["content"].startswith("/no_think")
 
 
 
@@ -790,7 +790,7 @@ def test_compute_node_runtime_readiness_smoke_completion_accepts_empty_qwen_thin
     assert diagnostics["api_v1_readiness_completion_smoke_result"] == "passed"
     assert diagnostics["api_v1_readiness_completion_smoke_shape"] == "api_v1_assistant_message"
     assert llm_runtime.completion_kwargs["max_tokens"] == 64
-    assert llm_runtime.completion_kwargs["messages"][-1]["content"].startswith("/no_think")
+    assert not llm_runtime.completion_kwargs["messages"][-1]["content"].startswith("/no_think")
 
 
 def test_compute_node_runtime_readiness_smoke_completion_empty_after_strip(monkeypatch):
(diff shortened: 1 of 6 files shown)


futuroptimist:

PR #1415, opened 2026-07-07
Enforce Qwen API v1 non-thinking, remove automatic `/no_think` injection, add Qwen non-thinking template fallback and hardened diagnostics

futuroptimist:
### Motivation

- Production Qwen API v1 readiness was failing in the render-then-plain-completion path and the UI lacked safe scalar diagnostics naming the exact rejected kwarg/method.
- The prior implementation relied on injecting a message-level `/no_think` control, which mutates user content and is unacceptable for the API-level non-reasoning invariant.
- A deterministic template fallback and stronger invariants are required so `enable_thinking=False` is authoritative, never stripped, retried away, or replaced by hidden message mutations.

### Description

- Completely removed automatic `/no_think` injection by making `RelayClient._api_v1_qwen_no_think_messages()` a no-op and ensuring `_api_v1_prepare_qwen_non_thinking_messages()` copies but does not mutate user messages; tests expecting injected `/no_think` were updated accordingly.
- Added `_render_qwen_api_v1_non_thinking_template(messages, llama, *, add_generation_prompt=True)` inside the subprocess worker code in `utils/llm/model_manager.py` to deterministically render Qwen ChatML with an empty `<think></think>` assistant block when needed, using the exact markers `<|im_start|>`/`<|im_end|>` and optionally prepending the Llama BOS token when present.
- Reworked `_render_chat_with_runtime_template(...)` to: (a) require `enable_thinking` to be exactly `False` when present and fail closed on invalid values, (b) require the hard-switch for Qwen API v1 non-thinking when the policy/provider indicates it, (c) try direct `apply_chat_template(..., enable_thinking=False)` but never retry by dropping `enable_thinking`, (d) prefer the GGUF/Jinja renderer but, if it cannot be trusted to honour `enable_thinking=False`, fall back to the deterministic Qwen template.
- Hardened the parent relay filter in `utils/networking/relay_client.py` so Qwen API v1 always includes `completion_kwargs["enable_thinking"] = False`, marks it as never-filtered, and fails closed with safe diagnostics (internal reason `runtime_qwen_non_thinking_hard_switch_unavailable`) if a child reports it rejected `enable_thinking`.
- Kept plain-completion attempts strictly bounded: `create_completion(prompt=..., max_tokens=...)`, `create_completion(..., max_tokens=...)`, and `llama(rendered_prompt, max_tokens=...)`; no unbounded fallbacks are attempted and rejected-kwarg diagnostics are surfaced safely.
- Expanded and propagated safe diagnostic fields across worker/relay/compute-node code to include the new plain-completion capability booleans and precise `rejected_generation_kwarg`, `attempted_generation_kwargs`, `attempted_plain_completion_methods`, `method`, and `generation_exception_category`, and updated readiness mapping to derive more specific readiness reasons when possible.

### Testing

- Ran focused unit suites and diagnostics validation: `python -m pytest tests/unit/test_model_manager.py -q`, `python -m pytest tests/unit/test_relay_client.py -q`, `python -m pytest tests/unit/test_compute_node_runtime.py -q`, and `python -m pytest tests/unit/test_desktop_compute_node_bridge.py -q`; all executed unit tests passed locally (unit suites green in this run).
- Exercised targeted model_manager worker-path tests that validate render fallback, enable_thinking enforcement, and plain-completion failure diagnostics; those tests passed after the changes and the worker fallback emits the exact empty `<think></think>` assistant generation prefix while never mutating user text.
- Verified Python syntax and runtime import checks via `python -m py_compile` during the edit cycle; no syntax errors remained.

Residuals & notes

- Automatic `/no_think` injection is removed; user-provided literal `/no_think` is preserved as ordinary user content by design and tests prove the difference.
- The implementation ensures `enable_thinking=False` is the authoritative, never-dropped hard switch for API v1 Qwen paths and surfaces safe scalar diagnostics when a runtime rejects it (the real rejected kwarg is surfaced as `rejected_generation_kwarg = "enable_thinking"` and the related `internal_reason`/`generation_exception_category` map to `runtime_qwen_non_thinking_hard_switch_unavailable`/`qwen_non_thinking_hard_switch_unavailable`).
- Unbounded generation fallbacks were explicitly removed; the code only attempts the bounded call shapes documented and fails closed with safe diagnostics if `max_tokens` is unsupported.

Verification commands run (high level)

- `python -m pytest tests/unit/test_model_manager.py -q` — passed
- `python -m pytest tests/unit/test_relay_client.py -q` — passed
- `python -m pytest tests/unit/test_compute_node_runtime.py -q` — passed
- `python -m pytest tests/unit/test_desktop_compute_node_bridge.py -q` — passed
- `python -m py_compile utils/llm/model_manager.py utils/networking/relay_client.py utils/compute_node_runtime.py` — passed

If you want, I can open a follow-up to run the full integration suites (`tests/integration/...`) in CI or attempt them here (they require a larger environment and longer runtime).

------
[Codex Task](https://chatgpt.com/codex/cloud/tasks/task_e_6a4c99d7c71c832fbf4a38801da9a1d7)

### DIFF
--- a/tests/unit/test_compute_node_runtime.py
+++ b/tests/unit/test_compute_node_runtime.py
@@ -747,7 +747,7 @@ def test_compute_node_runtime_readiness_smoke_completion_passes(monkeypatch):
     assert llm_runtime.completion_kwargs["max_tokens"] == 64
     assert "stream" not in llm_runtime.completion_kwargs
     assert "stop" not in llm_runtime.completion_kwargs
-    assert llm_runtime.completion_kwargs["messages"][-1]["content"].startswith("/no_think")
+    assert not llm_runtime.completion_kwargs["messages"][-1]["content"].startswith("/no_think")
 
 
 
@@ -790,7 +790,7 @@ def test_compute_node_runtime_readiness_smoke_completion_accepts_empty_qwen_thin
     assert diagnostics["api_v1_readiness_completion_smoke_result"] == "passed"
     assert diagnostics["api_v1_readiness_completion_smoke_shape"] == "api_v1_assistant_message"
     assert llm_runtime.completion_kwargs["max_tokens"] == 64
-    assert llm_runtime.completion_kwargs["messages"][-1]["content"].startswith("/no_think")
+    assert not llm_runtime.completion_kwargs["messages"][-1]["content"].startswith("/no_think")
 
 
 def test_compute_node_runtime_readiness_smoke_completion_empty_after_strip(monkeypatch):
--- a/tests/unit/test_model_manager.py
+++ b/tests/unit/test_model_manager.py
@@ -5854,7 +5854,7 @@ def test_subprocess_worker_plain_completion_helpers_cover_safe_shapes():
     assert classify_shape(TypeError("got an unexpected keyword argument 'stream'")) == 'unsupported_stream_kwarg'
     assert classify_shape(TypeError("got an unexpected keyword argument 'stop'")) == 'unsupported_stop_kwarg'
     assert classify_shape(TypeError("got an unexpected keyword argument 'temperature'")) == 'unexpected_kwarg'
-    assert classify_shape(RuntimeError("KV cache allocation failed")) == 'worker_exception'
+    assert classify_shape(RuntimeError("KV cache allocation failed")) == 'kv_cache_allocation'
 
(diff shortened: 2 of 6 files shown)


futuroptimist:

PR #1475, opened 2026-07-15
Fix desktop operator provisioning startup hang

futuroptimist:
### Motivation
- Packaged Windows startup could hang at “Worker state: starting” because dependency discovery ran before the managed site was prepended to import resolution, allowing blocking pip builds (and overlapping installers) to delay any startup event.
- Exact-pin llama-cpp runtime mismatches and unbounded/subprocess-buffered provisioning paths allowed long builds and fall-through repairs to leave the UI without an authoritative spawned/provisioning status or a responsive Stop operator.

### Description
- Make dependency discovery idempotent and read-only-aware by resolving the app-managed dependency target, prepending it into import resolution before `find_spec`, and exposing `ensure_desktop_python_dependencies(read_only=True)` for `model_bridge inspect` so `inspect` never mutates provisioned targets.  Also preserve `TOKEN_PLACE_DESKTOP_DEPENDENCY_TARGET` in `os.environ` when the target is used. (python: `desktop_runtime_setup.py`, `model_bridge.py`)
- Add a cross-process managed-site lock (`_ManagedSiteLock`) and re-check missing modules after acquiring the lock so a second packaged process with a populated managed site makes zero pip calls; install only allowlisted missing baseline modules rather than reinstalling the full baseline. (python: `desktop_runtime_setup.py`)
- Remove the accidental desktop bootstrap dependency on `packaging` and replace `packaging`-based version resolution with a strict stdlib-based exact match that allows exact public versions and local-build suffixes (e.g. `0.3.32` and `0.3.32+local`), rejecting prerelease/post/dev/malformed versions. (files: `requirements_desktop_runtime.txt`, `desktop_runtime_setup.py`)
- Replace blocking buffered `subprocess.run(capture_output=True)` for pip/source builds with a managed loop that drains pipes in small timeouts, emits safe provisioning heartbeats every ~5s, redacts commands in logs, and enforces existing timeouts by terminating the whole process tree on timeout/cancel. Add `--upgrade` to CUDA source-repair flows so the promoted target is replaced atomically. (python: `desktop_runtime_setup.py`)
- Emit an authoritative session-scoped provisioning status immediately after the bridge is spawned (Rust side) that includes the real operator log path, `running=true`, `registered=false`, `worker_alive=false`, `worker_state=provisioning`, and monotonic sequence fencing so the frontend clears its start latch and Stop remains enabled during provisioning. (Rust: `compute_node.rs`; frontend: `App.tsx`)
- Make `model_bridge inspect` use read-only dependency preflight and fall back to safe metadata instead of launching pip; preserve existing stdlib-before-third-party and repo-shim protections. (python: `model_bridge.py`)
- Add safe, bounded readiness diagnostics keys and display provisioning phase/elapsed/deadline in the UI while ensuring readiness diagnostics are allowlisted and sanitized. (frontend: `App.tsx`)
- Add regression unit tests that assert managed-site reuse, post-lock recheck behavior, bounded subprocess timeout/heartbeat behavior, and exact/local llama-cpp version matching. (tests: `tests/unit/test_desktop_runtime_setup.py`)

### Testing
- `python -m pip install -r config/requirements_codex_verification.txt` — PASS (installed verification deps)
- `python -m py_compile desktop-tauri/src-tauri/python/desktop_runtime_setup.py desktop-tauri/src-tauri/python/compute_node_bridge.py desktop-tauri/src-tauri/python/model_bridge.py` — PASS (compilation OK)
- `pytest -q tests/unit/test_desktop_runtime_setup.py -q` — PASS (102 passed)
- `pytest -q tests/unit/test_desktop_compute_node_bridge.py tests/unit/test_desktop_model_bridge.py -q` — PASS (147 passed)
- `pytest -q tests/integration/test_api_v1_desktop_bridge_e2ee.py -q` — PASS (10 passed; warnings only)
- `pytest -q tests/integration/test_desktop_qwen64k_packaged_runtime.py -q` — PASS (20 passed)
- `npm install` (desktop-tauri) — PASS (packages installed; npm audit reports unrelated vulnerabilities)
- `npm test` (desktop-tauri) — PASS (vitest: 55 passed)
- `pre-commit run --all-files` — PASS (hooks passed)
- `cd desktop-tauri/src-tauri && cargo check` — WARN (not completed here due to missing system `glib-2.0.pc` in this Linux container; run on a machine with system deps for a full Rust build check)
- `pytest -q desktop-tauri/scripts/test_packaged_operator_e2e.py -q` — WARN (collected no tests; this target is a packaging script/test harness and requires staged packaged artifacts and platform hardware to exercise)

Residual risks & follow-ups
- Hardware-validated Windows 11 CUDA checks must be executed on staging Windows machines (see staging validation note in this PR) because CI uses mocked runtimes; the code enforces fail-closed behavior for the Win11 Qwen64K tier and must be validated on real Windows+CUDA hardware.
- The Rust `cargo check` step was not completed in this container due to a missing system library; ensure CI/staging runs a full Rust toolchain check on the target builder with `glib`/pkg-config installed.

If you want, I can open a focused follow-up PR to harden Windows Job Object process-tree termination logic or add additional packaged-e2e harness wiring for Windows staging validation.

------
[Codex Task](https://chatgpt.com/codex/cloud/tasks/task_e_6a56f42eee24832f953f3ba6d9691082)

### DIFF
--- a/desktop-tauri/src-tauri/python/compute_node_bridge.py
+++ b/desktop-tauri/src-tauri/python/compute_node_bridge.py
@@ -780,7 +780,7 @@ def _normalize_relay_urls(*raw_relay_url_groups: Any) -> List[str]:
 def run(args: argparse.Namespace) -> int:
     bridge_session_id = _bridge_session_id_from_env()
     _reset_bridge_lifecycle_state(bridge_session_id)
-    status_sequence = 0
+    status_sequence = int(os.environ.get("TOKEN_PLACE_DESKTOP_EVENT_SEQUENCE_BASE", "0") or "0")
     emit_lock = threading.Lock()
 
     def emit_operator_event(payload: Dict[str, Any]) -> None:
--- a/desktop-tauri/src-tauri/python/desktop_runtime_setup.py
+++ b/desktop-tauri/src-tauri/python/desktop_runtime_setup.py
@@ -6,6 +6,7 @@ import importlib.util
 import json
 import os
 import platform as platform_module
+import re
 import subprocess
 import sys
 import time
@@ -102,6 +103,11 @@ GPU_RUNTIME_FATAL_ACTIONS = frozenset(
         "unavailable",
         "metal_install_failed",
         "metal_cpu_fallback",
+        "version_mismatch_failed",
+        "install_failed",
+        "install_timeout",
+        "install_cancelled",
+        "runtime_verification_failed",
     }
 )
 PIP_INSTALL_TIMEOUT_SECONDS = 300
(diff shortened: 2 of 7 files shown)


futuroptimist:

PR #1481, opened 2026-07-15
Fix macOS Qwen 64K runtime identity handoff

futuroptimist:
### Motivation

- Packaged macOS probes showed the bundled runtime was healthy but the compute-node/ModelManager handoff lost the authoritative llama_cpp binding because the raw `llama_module_path` was removed too early and got coerced to `"unknown"` down the consumer chain. 
- The redaction intended to protect path privacy prevented the subprocess facade (which has a concrete `__file__`) from matching the desktop probe, causing the Qwen 64K capability gate to fail closed on otherwise-valid macOS bundles. 
- The change restores a verifiable, path-free identity that preserves privacy while allowing an authoritative match between producer probe and consumer facade without launching reprobes (preserving Windows no-reprobe protection).

### Description

- Add a canonicalization+hash identity: compute `llama_module_identity = sha256:<64 hex>` from a canonical resolved module path (symlinks/`..`, case/separator normalization, Windows extended-prefix stripping) in `desktop_runtime_setup.py` and `utils/llm/model_manager.py`. 
- Carry `llama_module_identity` and a safe boolean `llama_module_path_present` in the private runtime probe payload (`TOKEN_PLACE_DESKTOP_RUNTIME_PROBE_JSON`) while continuing to omit raw paths from the public probe result; public output redacts both raw path and identity. 
- Make the compute bridge prefer the private probe env payload when assigning `ModelManager.desktop_runtime_probe` so the background warm-load/facade path receives the opaque identity without leaking it to logs. 
- Validate identity schema strictly and require an exact identity match (or canonical path agreement for legacy probes) in `_runtime_supports_qwen_yarn_rope()`, preserve fail-closed semantics for missing/malformed/mismatched identity, and avoid launching secondary native reprobes. 
- Fix diagnostic formatting in `_format_qwen_yarn_unsupported_diagnostics()` so False and empty-list values are serialized literally (not as `unknown`) while forcing raw module path diagnostics to remain `unknown` (privacy). 
- Add unit/regression tests exercising the full producer→consumer chain: authoritative identity handoff (no reprobe), missing/malformed/mismatched identity failure, canonicalization (symlinks/`..`), and payload redaction/privacy checks.

### Testing

- Installed focused test deps: `python -m pip install -r config/requirements_codex_verification.txt` (succeeded). 
- Ran unit test sweep: `python -m pytest tests/unit/test_desktop_runtime_setup.py tests/unit/test_desktop_compute_node_bridge.py tests/unit/test_model_manager.py -q` and observed all unit tests pass (551 passed). 
- Ran integration tests: `python -m pytest tests/integration/test_desktop_qwen64k_packaged_runtime.py tests/integration/test_macos_release_probe.py -q` and observed success (20 passed, 1 skipped on Linux). 
- Ran operator E2E and relay invariants: `python desktop-tauri/scripts/test_packaged_operator_e2e.py` and `python -m pytest tests/unit/test_crypto_helpers_api_v1_relay.py tests/unit/test_e2ee_relay_invariant.py -q` (succeeded). 
- Pre-commit and project checks: `pre-commit run --all-files` and `git diff --check` were executed and passed.

------
[Codex Task](https://chatgpt.com/codex/cloud/tasks/task_e_6a57fc405d08832f94f8fe49f7ab080a)

### DIFF
--- a/desktop-tauri/src-tauri/python/compute_node_bridge.py
+++ b/desktop-tauri/src-tauri/python/compute_node_bridge.py
@@ -954,9 +954,13 @@ def run(args: argparse.Namespace) -> int:
     context_profile = apply_context_profile(runtime.model_manager, args.context_tier)
     apply_compute_mode(runtime.model_manager, args.mode)
     try:
-        runtime.model_manager.desktop_runtime_probe = dict(runtime_setup)
+        private_probe = json.loads(os.getenv("TOKEN_PLACE_DESKTOP_RUNTIME_PROBE_JSON", "{}"))
+        runtime.model_manager.desktop_runtime_probe = dict(private_probe if isinstance(private_probe, dict) else runtime_setup)
     except Exception:
-        pass
+        try:
+            runtime.model_manager.desktop_runtime_probe = dict(runtime_setup)
+        except Exception:
+            pass
 
     warm_load_enabled = _env_enabled("TOKENPLACE_DESKTOP_WARM_LOAD", WARM_LOAD_DEFAULT)
     runtime_path = _runtime_path_from_env()
--- a/desktop-tauri/src-tauri/python/desktop_runtime_setup.py
+++ b/desktop-tauri/src-tauri/python/desktop_runtime_setup.py
@@ -2,8 +2,10 @@
 
 from __future__ import annotations
 
+import hashlib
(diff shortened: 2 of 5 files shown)


futuroptimist:

PR #1482, opened 2026-07-15
Restore opaque llama_cpp module identity handoff to fix macOS Qwen 64K capability gate

futuroptimist:
### Motivation

- Packaged macOS reported a healthy bundled runtime (CPython + llama-cpp-python 0.3.32 + Metal offload + YaRN support) but the Qwen 64K capability gate failed because the desktop probe redacted `llama_module_path` early and the compute-side coercion converted that missing path to `"unknown"`, so the background-thread `_SubprocessLlamaCppModule` facade could not be matched to the authoritative probe. 
- The intent is to preserve path privacy while restoring a verifiable, import-free binding between the producer (desktop probe) and consumer (ModelManager/_SubprocessLlamaCppModule) so the capability handoff succeeds on packaged macOS without reintroducing raw paths into logs or public payloads.

### Description

- Introduces an opaque, versioned, path-free `llama_module_identity` computed from a canonicalized resolved module path using a domain-separated SHA-256 digest and adds `llama_module_path_present` to record raw-path presence without exposing it (`desktop-tauri/src-tauri/python/desktop_runtime_setup.py`).
- Carries only the opaque identity and the safe presence boolean through the serialized probe payload and validates the identity strictly on the consumer side (`utils/llm/model_manager.py`), computing the same digest from the facade `__file__` and requiring an exact match to authorize the subprocess facade.
- Preserves fail-closed semantics and legacy behavior by: failing when identity is missing/malformed/mismatched, failing on identity/path disagreements, keeping concrete-path matching for legacy probes, and avoiding any secondary native reprobe that could hang on Windows.
- Fixes misleading diagnostics formatting so `False` and empty containers are rendered correctly, and extends tests to cover the producer/consumer contract, canonicalization, privacy, negative identity cases, legacy-path behavior, and the complete packaged-background warm-load capability chain (tests and locations updated under `tests/unit` and `tests/integration`).

### Testing

- Ran unit and integration suites and local e2e smoke used in validation with these commands: `python -m pytest tests/unit/test_desktop_runtime_setup.py tests/unit/test_desktop_compute_node_bridge.py tests/unit/test_model_manager.py -q`, `python -m pytest tests/integration/test_desktop_qwen64k_packaged_runtime.py tests/integration/test_macos_release_probe.py -q`, and `python desktop-tauri/scripts/test_packaged_operator_e2e.py`.
- Also ran relay/E2EE checks and project linters: `python -m pytest tests/unit/test_crypto_helpers_api_v1_relay.py tests/unit/test_e2ee_relay_invariant.py -q` and `pre-commit run --all-files`.
- All invoked automated tests passed in this environment and new unit tests added for identity/canonicalization/privacy/diagnostics passed.
- Notes and limits: the macOS real-app + mounted-DMG artifact probe is covered by the added integration tests but the packaged macOS artifact runtime validation requires running the release-mount validation on macOS runner as described in the validation checklist; this PR does not change any Windows provisioning behavior from the separate Windows-focused work in progress.

------
[Codex Task](https://chatgpt.com/codex/cloud/tasks/task_e_6a57fc474a60832fb3f7775783e9d65f)

### DIFF
--- a/desktop-tauri/src-tauri/python/desktop_runtime_setup.py
+++ b/desktop-tauri/src-tauri/python/desktop_runtime_setup.py
@@ -2,6 +2,7 @@
 
 from __future__ import annotations
 
+import hashlib
 import importlib.util
 import json
 import os
@@ -51,6 +52,25 @@ def _strip_windows_extended_path_prefix(path_text: str) -> str:
 
 def _safe_resolve_path(path_text: str | Path) -> Path:
     return Path(_strip_windows_extended_path_prefix(str(path_text))).resolve()
+
+
+def _canonical_llama_module_identity_path(path_text: str | Path) -> str:
+    return os.path.normcase(
+        os.path.normpath(str(_safe_resolve_path(path_text))).replace(os.sep, "/")
+    )
+
+
+def llama_module_identity_from_path(path_text):
+    if not path_text:
+        return None
+    try:
+        canonical = _canonical_llama_module_identity_path(path_text)
+    except (TypeError, ValueError, OSError):
+        return None
+    digest = hashlib.sha256(
+        ("token.place.llama_module_identity.v1\0" + canonical).encode("utf-8", "surrogatepass")
+    ).hexdigest()
+    return f"sha256:{digest}"
 
 
 @dataclass(frozen=True)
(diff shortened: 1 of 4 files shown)


futuroptimist:

PR #1488, opened 2026-07-17
Enforce API v1 message abuse ceiling in UTF-8 bytes (512 KiB) and add regression tests

futuroptimist:
### Motivation
- The structural API v1 guard previously used a 131,072 Python-character ceiling and rejected very large natural-language prompts before authoritative token admission could run, blocking valid ~242 KB prompts that fit a 64k Qwen context.
- We need a bounded pre-tokenization abuse/transport guard but it must be expressed in transport bytes (UTF-8) and be large enough to let legitimate large prompts reach the runtime tokenizer for exact admission.

### Description
- Raised the aggregate abuse ceiling and changed units by adding `
_API_V1_MAX_TOTAL_MESSAGE_UTF8_BYTES = 512 * 1024
` and keeping a backward-compatible alias `
_API_V1_MAX_TOTAL_REQUEST_CHARS = _API_V1_MAX_TOTAL_MESSAGE_UTF8_BYTES
` in `utils/networking/relay_client.py`.
- Switched validation to count incremental UTF-8 bytes instead of Python `len()` codepoints by returning `(char_count, utf8_bytes)` from `_api_v1_content_validation_size()` and summing `total_content_utf8_bytes` during `_validate_api_v1_chat_messages()` while preserving `total_content_chars` for compatibility.
- Added safe diagnostics and logging fields (`total_content_utf8_bytes`, `maximum_total_content_utf8_bytes`, `message_content_utf8_bytes`) and surfaced them on `compute_node_request_too_large` errors without exposing plaintext content, preserving relay-blind E2EE and privacy invariants.
- Left exact runtime rendering/tokenization as the sole authority for context admission (`compute_node_context_window_exceeded` remains the definitive rejection when tokens overflow the active tier).
- Added focused regression tests in `tests/unit/test_relay_client.py` that generate a deterministic 240–260 KiB ASCII/natural-language payload at runtime and exercise a lightweight Qwen-like runtime bridge to assert: completion succeeds on `64k-full` (55,229 prompt tokens + 512 output tokens admitted), fails with `compute_node_context_window_exceeded` on `8k-fast`, and that token-overflow cases continue to be rejected by token admission rather than the byte ceiling; tests also cover exact UTF-8-byte boundary and multibyte character behavior.

### Testing
- Installed focused verification deps with `pip install -r config/requirements_codex_verification.txt` (success).
- Ran unit tests: `python -m pytest tests/unit/test_relay_client.py -q` — all selected/additional tests passed (full suite: `313 passed`).
- Ran touch UI unit tests: `python -m pytest tests/unit/test_touch_ui.py -q` — `14 passed`.
- Ran integration tests: `python -m pytest tests/integration/test_api_v1_desktop_bridge_e2ee.py -q` — `10 passed`.
- Ran repository checks: `git diff --check` and `pre-commit run --all-files` — pre-commit hooks passed.

Verification summary: the change raises the abuse ceiling to `512 KiB` measured in aggregate UTF-8 bytes, preserves character-count diagnostics, adds explicit UTF-8 byte diagnostics in errors/logs, keeps runtime tokenization as the admission authority, preserves relay-blind E2EE and non-streaming API v1 invariants, and is covered by new regression tests which passed locally along with the existing test suite.

------
[Codex Task](https://chatgpt.com/codex/cloud/tasks/task_e_6a59a921ca54832f81d5b28d3af036cf)

### DIFF
--- a/tests/unit/test_relay_client.py
+++ b/tests/unit/test_relay_client.py
@@ -4564,8 +4564,134 @@ def test_api_v1_invalid_messages_are_rejected_before_worker(messages):
     assert (manager.worker_health, manager.recovery_count) == before
 
 
+
+
+def _deterministic_natural_payload(target_bytes=248 * 1024):
+    phrase = (
+        "This neutral validation sentence describes calm weather, ordinary numbers, "
+        "and predictable spacing for a deterministic prompt fixture. "
+    )
+    repeats = (target_bytes // len(phrase.encode("utf-8"))) + 1
+    payload = (phrase * repeats)[:target_bytes]
+    assert len(payload.encode("utf-8")) == target_bytes
+    return payload
+
+
+class _QwenAdmissionRuntime:
+    def __init__(self, prompt_tokens):
+        self.prompt_tokens = prompt_tokens
+        self.render_and_tokenize_calls = []
+        self.completion_calls = []
+
+    def render_and_tokenize_chat(self, messages, **kwargs):
+        self.render_and_tokenize_calls.append((messages, kwargs))
+        return {"prompt_tokens": self.prompt_tokens}
+
+    def create_chat_completion_from_rendered_prompt(self, messages, **kwargs):
+        self.completion_calls.append((messages, kwargs))
+        return {"choices": [{"message": {"role": "assistant", "content": "ok"}}]}
+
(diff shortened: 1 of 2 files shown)


futuroptimist:

PR #1493, opened 2026-07-17
Bundle deterministic Windows CUDA desktop runtime

futuroptimist:
### Motivation
- Packaged Windows builds were still invoking the host `py` launcher and performing a CUDA source build of `llama-cpp-python` at startup, creating a dependency on system Python, network/pip, CMake/MSVC/CUDA Toolkit, and local compilation instead of using an immutable bundled runtime. 
- The intent is to require a pinned, release-built Windows runtime (CPython 3.11 + llama-cpp-python 0.3.32 cu124) and to fail closed on invalid/missing runtime to guarantee releases never trigger `cuda_build` on end-user machines.

### Description
- Require and detect a bundled Windows runtime by preferring the resource `python-runtime/python.exe` and identifying it as `bundled-cpython-3.11-win-amd64-cu124`, implemented in `desktop-tauri/src-tauri/src/python_runtime.rs`. 
- Replace the Windows install plan so `llama_cpp` is resolved from the pinned cu124 wheel only (`llama_cpp_python-0.3.32-py3-none-win_amd64.whl`) and remove source-build flags like `--no-binary`, `CMAKE_ARGS=-DGGML_CUDA=on`, and `FORCE_CMAKE`, in `desktop-tauri/src-tauri/python/desktop_gpu_packaging.py`. 
- Make packaged Windows startup fail-closed before any pip/install/source-build attempts when the immutable bundled probe fails or reports non-CUDA, and disallow automatic source repair in bundled layouts unless explicit development opt-in is present, in `desktop-tauri/src-tauri/python/desktop_runtime_setup.py`. 
- Add a Windows embedded runtime manifest and provenance contract `desktop-tauri/src-tauri/python/embedded_python_runtime_manifest_windows_x86_64.json` plus a release artifact validator `desktop-tauri/scripts/validate_windows_release_artifact.py` and a prepare helper update `desktop-tauri/scripts/prepare_embedded_python_runtime.py` to support Windows runtime production preparation. 
- Add regression tests `tests/test_windows_packaged_cuda_runtime_contract.py` that assert the Windows plan is cu124 wheel-only, the manifest pins the wheel and DLL inventory, and a bundled runtime probe does not call pip or emit the `cuda_build` startup phase. 
- Update packaging metadata and versions from `0.1.1`/`0.1.0` to `0.1.2` for Tauri/npm/Cargo artifacts so `desktop-v0.1.2` artifacts report `0.1.2` consistently.

### Testing
- Installed verification deps with `pip install -r config/requirements_codex_verification.txt`, which completed successfully. 
- Ran the new unit/regression suite `python -m pytest -q tests/test_windows_packaged_cuda_runtime_contract.py` and all tests passed (3 passed). 
- Verified the embedded runtime prep helper help and manifest loader with `python desktop-tauri/scripts/prepare_embedded_python_runtime.py --target windows-x86_64 --cache-dir /tmp/... --help` and an import probe of `load_manifest(...)`, both succeeded. 
- Static checks: `python -m py_compile` on modified Python modules, `cargo fmt --manifest-path desktop-tauri/src-tauri/Cargo.toml --check`, and `git diff --check` passed; `cargo test python_runtime` could not be executed in this Linux container due to missing `glib-2.0.pc` (Tauri system dependency), and `pre-commit run --all-files` began successfully but was interrupted in this environment while long project checks were running. 

(Notes/observability) The root cause was that packaged Windows still fell back to a host interpreter and then followed a CUDA source-build plan; the patch makes the bundled runtime authoritative, pins the cu124 wheel and runtime artifacts by manifest+SHA, removes source-build plans for release Windows, and adds tests and artifact validators; real Windows/RTX smoke validation and installer extraction still require a Windows/NVIDIA release CI gate.

------
[Codex Task](https://chatgpt.com/codex/cloud/tasks/task_e_6a59c4d91168832f84aa77298e57ea2b)

### DIFF
--- a/desktop-tauri/package-lock.json
+++ b/desktop-tauri/package-lock.json
@@ -1,12 +1,12 @@
 {
   "name": "token-place-desktop-tauri",
-  "version": "0.1.1",
+  "version": "0.1.2",
   "lockfileVersion": 3,
   "requires": true,
   "packages": {
     "": {
       "name": "token-place-desktop-tauri",
-      "version": "0.1.1",
+      "version": "0.1.2",
       "dependencies": {
         "@tauri-apps/api": "^2.0.0",
         "@tauri-apps/plugin-dialog": "^2.0.0",
--- a/desktop-tauri/package.json
+++ b/desktop-tauri/package.json
@@ -1,7 +1,7 @@
 {
   "name": "token-place-desktop-tauri",
   "private": true,
-  "version": "0.1.1",
+  "version": "0.1.2",
   "type": "module",
   "scripts": {
     "dev": "vite",
--- a/desktop-tauri/scripts/prepare_embedded_python_runtime.py
+++ b/desktop-tauri/scripts/prepare_embedded_python_runtime.py
@@ -20,6 +20,7 @@ from pathlib import Path, PurePosixPath
 ROOT = Path(__file__).resolve().parents[1]
 SRC_TAURI = ROOT / "src-tauri"
 MANIFEST = SRC_TAURI / "python" / "embedded_python_runtime_manifest.json"
+WINDOWS_MANIFEST = SRC_TAURI / "python" / "embedded_python_runtime_manifest_windows_x86_64.json"
 OUTPUT = SRC_TAURI / "python-runtime"
 PROVENANCE = "embedded_python_runtime_provenance.json"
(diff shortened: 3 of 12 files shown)


futuroptimist:

PR #1514, opened 2026-07-22
Refine Claude permissions and tool restrictions

futuroptimist:
## Summary

Security-hardens the Claude Code GitHub integration while preserving owner-invoked, ad-hoc code-writing support.

This PR:

- Keeps ordinary `@claude` requests owner-only and disables Bash completely.
- Adds a separately approved `@claude-exec` path for fixed validation commands and workflow changes.
- Uses API-based, commit-signed changes instead of authenticated Git pushes.
- Rejects fork PRs and prevents checkout credentials from being persisted.
- Restricts conversational context to comments from `futuroptimist`.
- Does not enable Claude as an automatic PR reviewer.

Only `.github/workflows/claude.yml` is changed.

### Ordinary `@claude`

The normal path:

- Runs only after an explicit `@claude` mention from the repository owner.
- Explicitly excludes comments containing `@claude-exec`.
- Retains repository file-writing, issue, pull-request, and Actions-read permissions.
- Does not receive `workflows: write`.
- Uses `use_commit_signing: true` for GitHub API/MCP-based commits.
- Uses `persist-credentials: false`.
- Denies Bash through Claude permissions, CLI arguments, and a `PreToolUse` hook.
- Confines Claude file tools to paths resolving inside `GITHUB_WORKSPACE`.
- Rejects pull requests whose head repository differs from the base repository.

### Privileged `@claude-exec`

The privileged path:

- Runs only after an explicit `@claude-exec` mention from `futuroptimist` with `OWNER` association.
- Uses the protected `claude-exec` GitHub Environment.
- Requires the environment to contain at least one required reviewer.
- Uses only the environment-scoped `CLAUDE_CODE_OAUTH_TOKEN_EXEC` secret.
- May request `workflows: write` after environment approval.
- Uses API-based commit signing and does not persist checkout credentials.
- Rejects fork pull requests.
- Allows Bash only when the command exactly matches one immutable, zero-argument validation wrapper.

The fixed wrappers cover dependency compatibility, Python tests, frontend lint, the repository test suite, and an isolation smoke test. They are created outside the checkout and cannot be modified by repository code.

## Sandbox boundary

Claude-invoked validation commands run through Bubblewrap with:

- New user, PID, and network namespaces.
- A private `/proc`.
- Outbound networking disabled.
- A cleared environment reconstructed without GitHub, OAuth, OIDC, or other secret-bearing variables.
- Isolated `HOME` and `/tmp` directories.
- Read-only system and toolchain mounts.
- A writable repository checkout only where validation requires it.
- Per-command locked call counters and bounded execution time.
- No unsandboxed fallback if Bubblewrap or user-namespace isolation is unavailable.

The workflow handles the Ubuntu 24.04 AppArmor user-namespace restriction using the same ephemeral compatibility setting as the official Claude action, then verifies that Bubblewrap can execute inside the resulting namespace. A second preflight exercises the actual isolation wrapper before Claude receives access to it.

Missing project dependencies fail closed and defer validation to normal CI; the privileged path does not install dependencies or enable package-registry access from inside the sandbox.

## Required one-time repository setup

Before using `@claude-exec`:

1. Create a GitHub Environment named `claude-exec`.
2. Configure at least one required reviewer.
3. Decide deliberately whether the solo repository owner may self-approve deployments.
4. Disable administrator/environment-protection bypass where feasible.
5. Add `CLAUDE_CODE_OAUTH_TOKEN_EXEC` as an environment secret, not a repository secret.

The privileged workflow intentionally fails before invoking Claude if the environment is missing or lacks a required-reviewer rule.

## Security model and limitations

Repository files, diffs, issue and PR descriptions, and other content Claude reads remain untrusted and may contain prompt injection. Restricting context to comments by `futuroptimist` is defense in depth; reviewer feedback from other actors must be deliberately summarized or quoted in the owner’s request.

The normal path cannot execute repository code or modify workflows. The privileged path can execute only the fixed sandboxed wrappers after protected-environment approval. Prompt injection may still influence proposed source edits, so human review and standard CI remain required before merging Claude-authored changes.

## Validation

- `actionlint .github/workflows/claude.yml`
- Workflow YAML and both embedded settings objects parsed successfully.
- Guard, wrapper, and sandbox heredocs passed shell syntax validation.
- `git diff --check`
- Ordinary `@claude` verified to have no Bash or workflow-write grant.
- `@claude-exec` verified not to trigger the ordinary job.
- Exact wrapper matching rejects arguments, shell indirection, pipelines, redirection, and environment-prefixed commands.
- Workspace guards reject traversal and paths resolving outside the checkout.
- Bubblewrap bootstrap uses a valid mounted `/bin/true` probe and fails closed when isolation is unavailable.
- Standard latest-head PR CI covers repository regression testing; the issue-comment-triggered privileged path additionally performs its own runtime isolation preflight before Claude starts.

### DIFF
--- a/.github/workflows/claude.yml
+++ b/.github/workflows/claude.yml
@@ -15,17 +15,22 @@ jobs:
     if: |
       (github.event_name == 'issue_comment' &&
         contains(github.event.comment.body, '@claude') &&
-        contains(fromJSON('["OWNER","MEMBER","COLLABORATOR"]'), github.event.comment.author_association)) ||
+        !contains(github.event.comment.body, '@claude-exec') &&
+        github.event.comment.author_association == 'OWNER') ||
       (github.event_name == 'pull_request_review_comment' &&
         contains(github.event.comment.body, '@claude') &&
-        contains(fromJSON('["OWNER","MEMBER","COLLABORATOR"]'), github.event.comment.author_association)) ||
+        !contains(github.event.comment.body, '@claude-exec') &&
+        github.event.comment.author_association == 'OWNER') ||
       (github.event_name == 'pull_request_review' &&
         contains(github.event.review.body, '@claude') &&
-        contains(fromJSON('["OWNER","MEMBER","COLLABORATOR"]'), github.event.review.author_association)) ||
+        !contains(github.event.review.body, '@claude-exec') &&
+        github.event.review.author_association == 'OWNER') ||
       (github.event_name == 'issues' &&
         (contains(github.event.issue.body, '@claude') ||
          contains(github.event.issue.title, '@claude')) &&
-        contains(fromJSON('["OWNER","MEMBER","COLLABORATOR"]'), github.event.issue.author_association))
+        !contains(github.event.issue.body, '@claude-exec') &&
+        !contains(github.event.issue.title, '@claude-exec') &&
+        github.event.issue.author_association == 'OWNER')
 
     runs-on: ubuntu-latest
     timeout-minutes: 45
@@ -38,9 +43,8 @@ jobs:
       id-token: write
 
     steps:
-      # Claude may execute repository code, so only permit PR tasks whose head
-      # branch belongs to this repository. Trusted users may still review fork
-      # PRs normally, but @claude executable tasks on those PRs fail closed.
+      # Claude receives untrusted repository and PR/comment content. Keep the
+      # normal path default-deny for executable tools and fail closed for forks.
       - name: Reject fork pull requests
         if: |
           (github.event_name == 'issue_comment' && github.event.issue.pull_request) ||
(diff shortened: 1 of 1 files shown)


futuroptimist:
@codex Please revise PR #1514 from its current head to address all four unresolved security-review threads while preserving useful, ad-hoc Claude code-writing capabilities.

The current `--allowedTools "Bash"` configuration must not remain. Implement the two-tier security design below, limiting the code diff to `.github/workflows/claude.yml` unless an additional workflow file is demonstrably necessary. Update the PR description as part of the task.

Primary threat model:
- Repository files, issue/PR bodies, and comments may contain prompt injection.
- A trusted owner invoking Claude does not make all content Claude reads trustworthy.
- The normal Claude process receives an Anthropic OAuth credential and a short-lived repository-scoped GitHub credential.
- Interpreters and test/build commands execute arbitrary repository-controlled code; `python`, `pytest`, `npm`, `npx`, `cargo`, and `pre-commit` are not inherently safe.
- Fail closed whenever a requested security boundary is unavailable.

Implement the following.

1. Restore default-deny executable tools

Remove:

- bare `--allowedTools "Bash"`; and
- the incomplete `--disallowedTools` Git denylist.

The normal `@claude` path must not add any general Bash capability. Preserve the action’s built-in `acceptEdits` behavior and built-in GitHub MCP/file tools.

Do not allow broad interpreter or package-manager patterns such as:

- `Bash(python:*)`
- `Bash(pytest:*)`
- `Bash(npm:*)`
- `Bash(npx:*)`
- `Bash(cargo:*)`
- `Bash(pre-commit:*)`
- arbitrary wrapper arguments

Anything not explicitly allowed must remain unavailable in the headless session.

2. Treat repository test/build commands as untrusted code

Do not describe the above commands as “safe.”

If Claude needs test/build execution, expose only fixed validation entry points created by the trusted workflow definition outside the repository checkout. Repository-controlled scripts, package metadata, hooks, tests, and build files must execute only inside the isolation boundary described below.

Do not install dependencies or access package registries from inside the isolated Claude-invoked validation commands. Missing dependencies should produce a clear, fail-closed result and defer to ordinary CI.

3. Use API-based commits and keep credentials out of Git configuration

Set:

`use_commit_signing: true`

Verify against the official `anthropics/claude-code-action` source that this selects the GitHub API/MCP commit path rather than normal authenticated Git pushes.

For every checkout used by a Claude job, set:

`persist-credentials: false`

Do not expose or allow raw `git add`, `git commit`, or `git push` commands. Do not place either the workflow token or Claude GitHub App token in `.git/config` or the `origin` URL.

The official action does contain its own guarded push helper, so do not claim that helper is nonexistent. With API commit signing enabled, however, do not depend on ordinary Git pushing at all.

4. Remove workflow-write access from normal `@claude`

The normal `@claude` job may retain:

- `contents: write`
- `pull-requests: write`
- `issues: write`
- `actions: read`
- `id-token: write`

Its action-level `additional_permissions` may retain only:

- `actions: read`

Remove `workflows: write` from the normal path.

Normal Claude requests may edit ordinary repository files and commit through the API, but must not be able to commit changes under `.github/workflows/**`.

Workflow modifications must be available only through the separately approved `@claude-exec` path below.

5. Add fixed, isolated validation entry points

Create fixed validation wrappers during the privileged job using literal workflow-controlled shell content and store them under `${{ runner.temp }}` or another path outside `${{ github.workspace }}`.

Requirements:

- Claude cannot edit or replace the wrappers.
- Each wrapper has one fixed purpose and accepts no user/model-controlled arguments.
- Derive the useful validation commands from the repository’s existing CI configuration.
- Prefer separate exact wrappers for applicable Python, frontend, Rust, or repository lint validation rather than one argument-driven dispatcher.
- Apply sensible per-script call caps if the current action supports `CLAUDE_CODE_SCRIPT_CAPS`.
- Allow Claude to invoke only the exact wrapper paths.
- Do not permit arbitrary interpreters directly.

Before repository-controlled code runs, the wrapper must enter an OS-enforced Linux isolation boundary that:

- clears the environment and explicitly reconstructs only non-secret variables needed by the toolchain;
- excludes `CLAUDE_CODE_OAUTH_TOKEN`, GitHub tokens, OIDC request variables, and every other secret-bearing variable;
- creates a new PID namespace with a private `/proc`, preventing inspection of the secret-bearing parent process;
- disables outbound network access;
- exposes only the minimum read-only system/toolchain paths;
- exposes the repository workspace as writable only where validation genuinely requires it;
- hides runner temporary directories, credentials, Git configuration outside the checkout, SSH material, and unrelated home-directory state;
- uses a temporary isolated `HOME` and `/tmp`;
- has a bounded execution timeout; and
- fails rather than falling back to unsandboxed execution if isolation cannot be established.

Use an appropriate supported mechanism such as Bubblewrap on `ubuntu-latest`. Install/prepare it in a fixed workflow step before Claude executes. Account for Ubuntu’s user-namespace restrictions where necessary.

Do not stack incompatible sandbox layers. If the current Claude action’s officially supported subprocess isolation can provide the required clean environment, PID isolation, and network denial for trusted-user mode, verify that behavior from the exact action source/version and use it. Otherwise implement the immutable outer wrapper explicitly.

6. Split ordinary and privileged invocations

Keep a normal `@claude` job and add a distinct privileged `@claude-exec` job.

Normal job:

- triggers on an explicit `@claude` mention;
- explicitly excludes comments containing `@claude-exec`, because that string also contains `@claude`;
- is owner-only for this repository;
- has no general Bash;
- has no `workflows: write`;
- uses API commit signing;
- keeps the existing same-repository/fork rejection;
- uses `persist-credentials: false`.

Privileged job:

- triggers only on an explicit `@claude-exec` mention;
- is restricted to repository owner `futuroptimist`/`OWNER`;
- keeps the same fork rejection and may never execute against a fork PR;
- declares `environment: claude-exec`;
- uses an environment-only secret named `CLAUDE_CODE_OAUTH_TOKEN_EXEC`;
- must not fall back to the normal repository OAuth secret;
- uses API commit signing and `persist-credentials: false`;
- may request `workflows: write` through the action only in this job;
- may invoke only the exact isolated wrappers from step 5—still never bare Bash.

Add a fixed preflight before the privileged Claude step that queries the `claude-exec` environment configuration and verifies that a required-reviewer protection rule exists. Fail before invoking Claude if the environment is missing or unprotected. Confirm the exact GitHub API response schema and required read permission rather than guessing.

The PR description must explain the one-time repository setup:

- create the `claude-exec` environment;
- configure at least one required reviewer;
- decide deliberately whether the solo owner may self-approve;
- disable administrator bypass if feasible;
- store the OAuth credential only as the environment secret `CLAUDE_CODE_OAUTH_TOKEN_EXEC`;
- understand that `@claude-exec` remains unavailable/fails closed until this setup is complete.

Ensure an `@claude-exec` comment cannot accidentally start both jobs.

7. Reduce untrusted context and repository-controlled configuration

For both action invocations, set:

`include_comments_by_actor: "futuroptimist"`

Only the owner’s comments should be supplied as conversational context. Document that reviewer comments from other actors must be summarized or quoted deliberately in the owner’s trigger comment.

Also prevent repository-controlled Claude configuration from silently broadening permissions or launching an untrusted MCP server:

- use `--setting-sources user` so project/local settings are not loaded;
- use `--strict-mcp-config` so only the action’s explicitly supplied GitHub MCP configuration is used;
- verify these flags compose correctly with the action-generated MCP configuration and do not break its built-in comment, CI, or commit tools.

This filtering does not make issue bodies, PR descriptions, diffs, `CLAUDE.md`, or repository source trustworthy. Treat it as defense in depth, with the permission and sandbox boundaries remaining authoritative.

Preserve existing protections:

- explicit mention-only triggers;
- no automatic reviewer trigger;
- owner/member/collaborator association checks where still relevant, tightened to owner-only for these two paths;
- same-repository PR verification;
- `set -euo pipefail`;
- full checkout history/tags where needed;
- 45-minute job timeout;
- `show_full_output` remaining false;
- no Actions debug/full transcript output.

PR description

Replace the current placeholder description with an accurate security and behavior summary covering:

- normal `@claude`;
- privileged `@claude-exec`;
- the protected-environment setup;
- API-based commits;
- why arbitrary Bash/interpreters remain disallowed;
- the immutable no-network validation wrappers;
- owner-only invocation/context;
- fork rejection;
- standard versus privileged token permissions;
- limitations and residual prompt-injection risk;
- confirmation that Claude is not an automatic reviewer.

Validation

- Validate YAML syntax and all GitHub Actions expressions.
- Run `actionlint` or the repository’s applicable workflow linting.
- Run `git diff --check`.
- Verify no job contains bare `Bash`, broad interpreter/package-manager Bash patterns, or a denylist masquerading as a security boundary.
- Verify normal `@claude` cannot request workflow-write access.
- Verify checkout credentials are not persisted.
- Verify `@claude-exec` cannot trigger the normal job.
- Verify the privileged path fails closed for forks, non-owner actors, a missing/unprotected environment, unavailable sandboxing, missing dependencies, and attempted wrapper arguments.
- Verify the isolated command environment contains no token/secret/OIDC variables and has no outbound network path.
- Confirm the code diff remains limited to the Claude workflow file(s).
- Commit and push the changes to this PR branch.

In the final report, map every Greptile, Copilot, and Codex review thread to its resolution. Explicitly distinguish the valid unrestricted-Bash findings from Greptile’s uncertainty about the guarded push helper, which does exist in the official action source. Resolve or mark addressed threads where tooling permits.

futuroptimist:
@claude

Scope Lock
- Modify only `.github/workflows/claude.yml`.
- Keep the diff minimal and preserve owner-only invocation, fork rejection, protected-environment gating, separate OAuth secrets, API commit signing, checkout credential hardening, strict MCP configuration, and ordinary `@claude` being unable to modify workflows.
- Do not add broad Bash/interpreter/package-manager permissions.
- Do not modify PR metadata or resolve review threads.

### Reviewer comment resolution

Greptile, Copilot, and Codex correctly objected to unrestricted Bash on the earlier commit. The latest commit removed bare `Bash`, but the remaining security boundary is still incomplete: Claude Code `allowedTools` entries are pre-approvals rather than an exclusive allowlist, and read-only Bash commands may still run. Greptile’s repeated latest-head warning is therefore directionally valid even though arbitrary commands such as `curl` are not automatically approved.

Implement an authoritative boundary that:

- completely removes Bash from ordinary `@claude`;
- allows privileged `@claude-exec` to invoke only the exact immutable, zero-argument wrapper paths;
- rejects every other Bash spelling, including read-only commands, shell indirection, quoting tricks, environment prefixes, redirection, pipelines, and wrapper arguments;
- prevents Claude’s file-reading tools from reading `/proc`, `/sys`, `/dev`, runner temporary/configuration paths, or anything resolving outside `GITHUB_WORKSPACE`;
- adds a concise durable comment explaining that the guard is necessary because `allowedTools` alone is not an exclusive allowlist.

Preserve `use_commit_signing: true`; do not reintroduce raw Git push helpers.

### Implementation

1. Create workflow-controlled guard/settings files outside the checkout under `${{ runner.temp }}`, make them non-writable by Claude, and load them through a mechanism supported by the exact `anthropics/claude-code-action@v1` implementation. Prefer a fail-closed `PreToolUse` hook that parses tool-call JSON without `eval`.
2. For normal `@claude`, explicitly disallow the Bash tool and restrict built-in file reads to the checked-out workspace, including realpath/symlink validation.
3. For `@claude-exec`, have the immutable hook allow Bash only when the command is exactly one generated wrapper path with no arguments or shell syntax. Deny all other Bash calls before execution.
4. Set `trigger_phrase: "@claude-exec"` on the privileged action invocation so the action extracts the request after the complete trigger instead of interpreting it using the default `@claude` prefix.
5. Replace the current newline-delimited `CLAUDE_CODE_SCRIPT_CAPS` value. The documented format is JSON, not `path:count` lines. Verify whether the exact action version enforces caps for this owner-triggered mode. Use valid JSON if it does, and enforce the same limits independently inside the immutable wrapper layer with atomic/locked counters outside the workspace so the caps remain authoritative.
6. Make Bubblewrap startup work fail-closed on `ubuntu-latest`, including Ubuntu’s `kernel.apparmor_restrict_unprivileged_userns` behavior. Follow the current action’s supported setup where applicable and run the exact sandbox command rather than relying only on `kernel.unprivileged_userns_clone`.
7. Add a pre-Claude smoke-test step that invokes the generated `env-network-check` wrapper and proves the sandbox starts, receives no token/secret/OIDC variables, has no outbound network, and has isolated process/filesystem state. Ensure commands use binaries actually mounted and available inside the sandbox. Missing project dependencies must fail clearly and defer to CI; do not install packages or enable network access from a secret-bearing execution path.
8. Keep wrapper files, guard files, counters, temporary HOME, and temporary directories outside `GITHUB_WORKSPACE`; ensure repository code cannot modify them.

### Verification

- Run `actionlint .github/workflows/claude.yml`.
- Run the repository’s YAML/workflow validation and `git diff --check`.
- Exercise the guard directly with representative tool-call JSON:
  - allow each exact wrapper invocation;
  - reject wrapper arguments;
  - reject `cat /proc/self/environ`, `env`, `bash -c`, pipelines, redirection, command substitution, alternate quoting, and environment-prefixed commands;
  - reject reads outside the workspace and workspace symlinks resolving outside it.
- Validate any `CLAUDE_CODE_SCRIPT_CAPS` value with `jq -e` and test that exceeding every wrapper cap fails.
- Verify ordinary `@claude` has no Bash and no `workflows: write`.
- Verify `@claude-exec` cannot start the ordinary job and the privileged action uses the exact `@claude-exec` trigger phrase.
- Verify the privileged path still fails closed for forks, non-owner actors, an unprotected environment, unavailable sandboxing, missing dependencies, and wrapper arguments.
- Confirm the final diff remains limited to `.github/workflows/claude.yml`, then commit and push it to this PR branch.

futuroptimist:
@codex

Scope Lock
- Modify only `.github/workflows/claude.yml`.
- Keep the diff minimal and preserve the current owner/fork/environment guards, exact-wrapper Bash hook, ordinary-path Bash denial, workspace path confinement, API commit signing, isolated OAuth credentials, call counters, and no-network execution.
- Do not modify PR metadata or review-thread state.

### Reviewer comment resolution

Greptile’s latest review of `fbba6145` confirms that the earlier unrestricted-Bash findings from Greptile, Copilot, and Codex are addressed. Preserve those protections.

Two latest-head implementation issues remain:

- The Bubblewrap bootstrap does not yet make `@claude-exec` operational on `ubuntu-latest`. Ubuntu 24.04 enables `kernel.apparmor_restrict_unprivileged_userns=1`, while the workflow only runs a probe that will fail under that restriction. The probe also launches `true` without mounting a filesystem containing that executable.
- The `_sandbox` comment claims `CLAUDE_CODE_SCRIPT_CAPS` is undocumented and necessarily a no-op. The current action security documentation does document this JSON environment control. Retaining immutable local `flock` counters is still the stronger, version-independent design, but the rationale must be accurate.

### Implementation

1. Harden `Install Bubblewrap sandbox runtime` for the current Ubuntu 24.04 GitHub-hosted runner:
   - After installing Bubblewrap, inspect `kernel.unprivileged_userns_clone` when present and fail closed if user namespaces are unavailable.
   - When `/proc/sys/kernel/apparmor_restrict_unprivileged_userns` exists and is `1`, apply the compatibility setup used by the current official Claude action before probing Bubblewrap. On this ephemeral runner, set it to `0` with `sudo sysctl -w`; do not persist the setting.
   - Run a valid probe whose executable exists inside the temporary namespace, such as a read-only root bind followed by `/bin/true`.
   - Print useful diagnostic values on failure and never fall back to unsandboxed execution.
2. Keep the existing `Verify sandbox isolation before invoking Claude` step as the exact end-to-end check of the real `_sandbox` command.
3. Make `env-network-check` use an explicitly mounted runner binary such as `/usr/bin/python3`, so its result cannot depend on a tool-cache-only `python` path.
4. Correct the `CLAUDE_CODE_SCRIPT_CAPS` comment: acknowledge the documented JSON control, explain that this workflow intentionally uses immutable local `flock` counters as the authoritative enforcement layer, and do not restore the former invalid newline-delimited value.
5. Preserve the current cap values and exact-wrapper-only hook behavior.

### Verification

- Run `actionlint .github/workflows/claude.yml`.
- Parse the workflow YAML and both embedded settings JSON objects.
- Run `git diff --check`.
- Run `bash -n` or equivalent against the extracted guard, wrapper, and `_sandbox` heredocs.
- On Ubuntu 24.04, execute the corrected Bubblewrap bootstrap probe and the generated `env-network-check` wrapper.
- Verify the wrapper sees no token/secret/OIDC variables and cannot reach the network.
- Verify unavailable user namespaces or a failed isolation probe terminates the job before the Claude action.
- Confirm ordinary `@claude` remains Bash-free and `@claude-exec` still accepts only exact, zero-argument wrapper commands.
- Confirm the final diff remains limited to `.github/workflows/claude.yml`, then commit and push it to this PR branch.

